PR #23, opened 2015-04-01
Add auto-detecting CPU  on configure stage.

xianyi:
e.g.  /Path_to_BLIS/configure auto

Now, it only support detecting x86 and ARM CPUs.


fgvanzee:
Xianyi, thanks again for this work.

PR #28, opened 2015-05-24
Adding support for building on OS X

poulson:
In addition to creating a new profile, `sandybridge_osx_clang`, I have finished working around the missing `pthread_barrier` functionality and generalized the build system to support more general shared and static library suffixes (e.g., .dylib for shared libraries on OS X). Lastly, I enabled shared-library builds by default in each of the configurations.


jeffhammond:
This looks good to me but I have not tested.

@fgvanzee Would you like me test locally?  I recall you may be a "serious programmer" (Linux laptop person).

Maratyszcza:
Is the new osx configuration only due to `BLIS_ENABLE_PTHREADS` vs `BLIS_ENABLE_OPENMP`?

poulson:
@Maratyszcza I believe that was the only change I made to ``bli_config.h`` (most changes were elsewhere)

Maratyszcza:
@fgvanzee Maybe multithreading model should be an argument to `./configure` instead of an option in configuration file? After all, it is not specific to CPU microarchitecture

poulson:
@Maratyszcza Agreed; in this case, it's a matter of compiler support.

jeffhammond:
Is this still an issue?  I can't cleanly merge this PR with the latest master and I wonder if it needs to be sorted out or not.

devinamatthews:
Is the only issue that it doesn't work "out of the box" with Apple's version of gcc/llvm (due do missing pthread_barrier)? If so we have a non-pthread barrier already that could be hooked up easily. The so/dyld stuff could be helpful but the linker manages to figure out what you mean with .so as far as I can tell. I **don't** think the sandybridge_osx_clang configuration stuff is necessary with various new configure options.

On the other hand, if we are fine not supporting Apple's gimped compilers then I think there is nothing to do at this point.

jeffhammond:
It is worth adding the workaround for missing `pthread_barrier` since that is not POSIX anyways.

devinamatthews:
@tlrmchlsmth I think you have the skillz to pay the billz on this one.

tlrmchlsmth:
Actually I already fixed the pthread barrier thing but I didn't know this issue was here.
The default right now is to not use pthread barriers, and if you want to use them, you have to define BLIS_USE_PTHREAD_BARRIER.

jeffhammond:
I guess it can be resolved although it's good to setup Mac builds in Travis
to validate this and other changes.

If Mac is tested in Travis already, I apologize. I'm not using a client
that makes checking that trivial.


-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

devinamatthews:
Obsoleted by #108.

PR #36, opened 2015-12-11
Fix dgemm performance bug on sandy bridge

xianyi:
Intel Xeon E5-2680 (2.7GHz)
Old dgemm single thread perf: 24.xx GFlops
New dgemm single thread perf: 25.xx GFlops


fgvanzee:
Xianyi, thanks for this pull request. I have two quick questions:

1. I see that you relaxed the starting address and ldim alignment constraints. This makes sense, as I understand it, because AVX no longer requires alignment in order for aligned loads to execute, ONLY for them to perform optimally (ie: if used with unaligned memory, they will work fine, but just run slower). Please confirm that this was your thinking.

2. I noticed that you changed the `vmovapd` instructions to `vmovups`. I think I read somewhere that the `s` instructions are shorter and thus easier to decode than the `d` instructions. But, if `vmovapd`/`vmovaps` work fine on unaligned memory (just more slowly), why use `vmovups` instead? Is it once again a shorter instruction?

xianyi:
The `vmovapd` doesn't work on the unaligned address. I meet a SEGFAULT when `vmovapd` read unaligned address on sandy bridge.

Thus, I try unaligned accessing instructions, e.g. `vmovups` or `vmovupd`. Because of shorter instruction, I choose `vmovups`.

fgvanzee:
Xianyi, thanks for your comments. It appears you are right vis-a-vis `vmovapd` on unaligned data. I once remember reading about the relaxed AVX alignment constraints [here](https://software.intel.com/en-us/articles/introduction-to-intel-advanced-vector-extensions). It turns out that I missed a key passage:

> Intel® AVX has relaxed some memory alignment requirements, so now Intel AVX by default
> allows unaligned access; however, this access may come at a performance slowdown, so
> the old rule of designing your data to be memory aligned is still good practice (16-byte
> aligned for 128-bit access and 32-byte aligned for 256-bit access). **The main exceptions
> are the VEX-extended versions of the SSE instructions that explicitly required
> memory-aligned data: These instructions still require aligned data.**

(Emphasis added.)

So it sounds like, if the instruction has origins in SSE, and already encoded the aligned or unaligned affinity in the name, as `movapd`/`movupd` did, then we must continue to honor those semantics.

I assume that using `vmovupd` on aligned data would yield the expected higher performance that is guaranteed by `vmovapd`. Is this also your understanding?

Maratyszcza:
@xianyi `V*PD` instructions have the same length as `V*PS`. The `*PS` instructions in SSE were shorter than `*PD` variants because they didn't need an extra prefix, but in AVX instructions the legacy prefixes are encoded in constant-width field of VEX prefix.

xianyi:
@Maratyszcza , thank you for pointing out.

fgvanzee:
@Maratyszcza Thanks for chiming in.

Okay, so to summarize:
- Since the instruction lengths in AVX are the same, there is no decoding benefit to using `vmov[au]ps` instructions over the corresponding `vmov[au]pd`.
- We must use `vmovup[ds]` instructions if unaligned data is a possibility.

Do either of you know if the performance of `vmovup[ds]` tracks that of `vmovap[ds]` when executed on aligned data?

jeffhammond:
Should this PR be closed in lieu of #86 or does it still need to be addressed?

(In case it isn't obvious, I am looking through all the open PRs to see if they can be resolved)

devinamatthews:
This is fixed and can be closed.

PR #45, opened 2016-03-03
Use clock_gettime(CLOCK_MONOTONIC) and mach_absolute_time instead of gettimeofday

devinamatthews:
Since `gettimeofday` may or may not be high precision, is obsoleted by POSIX.1-2008, and is affected by NTP hijinks, I switched to `clock_gettime(CLOCK_MONOTONIC)` for Linux et al., and `mach_absolute_time` for OSX since `clock_gettime` is not supported there. I also updated the OS detection in bli_system.h using recommended preprocessor macros from [here](https://sourceforge.net/p/predef/wiki/OperatingSystems).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,6 +160,16 @@ endif
 
 
 #
+# --- Append OS-specific libraries to LDFLAGS ----------------------------------
+#
+
+ifeq ($(OS_NAME),Linux)
+LDFLAGS += -lrt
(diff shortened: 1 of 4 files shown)


fgvanzee:
Thanks for this patch, Devin. I've been meaning to upgrade to `clock_gettime()` for a long time.

PR #53, opened 2016-03-26
Add AVX implementation for sdotxv, ddotxv, daxpy

bluss:
Relatively simple dotv impls. These improve dotv especially when the data already fits in cache.

Is this a good approach? Is there a more general strategy that can be used to write these?

At least, this impl should be included into other configurations, at least haswell. What's the best way to have them share it? Right now they link kernel to avx and avx2 respectively.

Is pseudonymous authorship ok, and is there anything about contributing that I have missed? Thank you.

----

Edits: PR has been updated to focus on dotxv since first posting.

### DIFF
--- a/config/sandybridge/bli_kernel.h
+++ b/config/sandybridge/bli_kernel.h
@@ -124,6 +124,7 @@
 
 // -- dotv --
 
+#define BLIS_SDOTV_KERNEL       bli_sdotv_opt_var1
 #define BLIS_DDOTV_KERNEL       bli_ddotv_opt_var1
 
 // -- dotxv --
--- a/config/sandybridge/bli_kernel.h
+++ b/config/sandybridge/bli_kernel.h
@@ -120,6 +120,8 @@
 
 // -- axpyv --
 
+#define BLIS_DAXPYV_KERNEL      bli_daxpyv_opt_var1
+
 // -- copyv --
 
 // -- dotv --
--- a/config/sandybridge/bli_kernel.h
+++ b/config/sandybridge/bli_kernel.h
@@ -124,6 +124,8 @@
 
 // -- dotv --
 
+#define BLIS_DDOTV_KERNEL       bli_ddotv_opt_var1
+
 // -- dotxv --
 
 // -- invertv --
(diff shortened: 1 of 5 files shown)


bluss:
Another question, my code is adapted from a dotxf core2-sse3 kernel. This code (my code in the PR) now overuses restrict I think, should it not use restrict only on the pointer we write through? ddotv should allow being called with the same vector in both input arguments.

bluss:
These two are the issue that travis fails on I guess?

```
blis_ddotaxpyv_ncc_ccc             400    1.378   1.14e-13   FAILURE
blis_ddotaxpyv_ccn_ccc             400    1.359   1.42e-13   FAILURE
```

How do I run these locally. I had make test pass.

bluss:
I resolved the test running, (`make BLIS_ENABLE_TEST_OUTPUT=yes test` is the short of it).

Can it be that the AVX ddotv implementation is more accurate that the reference ddotv, since the AVX impl splits the sum over 16 accumulators?

bluss:
I pushed a way to resolve the test failure: implement dotxv instead.

I propose having the reference dotv call the DOTXV kernel if it exists. That way you don't have to implement both. I realize that might be a large change, and will maybe not be accepted.

fgvanzee:
bluss, thanks for your contributions. I'm approaching a very large commit that involves a code reorganization, including to the kernels directory (which has needed cleanup for a long, long time). I'm going to ask that you hold onto these new kernels for the time being and resubmit the pull request against the new code layout once it is available. I will send out a message to blis-devel when the commit is ready. (Sorry for the inconvenience.)

bluss:
Ok! Sounds good.

jeffhammond:
So has this attempted contribution been abandoned?

devinamatthews:
@bluss if you update the PR based on the current `master`, I imagine @fgvanzee will accept it. The "very large commit" that he mentions a couple comments up is complete (537a1f4f85ce1aa008901857cb3182e6b4546d7f) and he new code is stable.

bluss:
Yes, BLIS is super interesting (thanks for all the papers) but it turned out to not be the best fit for my project. Feel free to use my code.

fgvanzee:
Thanks @bluss, we understand. We appreciate your contributions nonetheless. (We also have another contributor who plans to add some level-1v kernels for AVX in the near future.)

PR #59, opened 2016-03-31
Fix testsuite makefile

devinamatthews:
Creates a new file, common.mk, which is included by both Makefile and testsuite/Makefile to determine build settings.


fgvanzee:
Thanks for this fix, Devin.

devinamatthews:
You're welcome. I'll add testsuite/Makefile to my testing regimen for 
the future.

PR #66, opened 2016-04-19
Add configure options and generate bli_config.h automatically.

### LINKED ISSUE #56: support LLVM OpenMP
`OpenMP is not supported with Clang` in the build system has been false [http://blog.llvm.org/2015/05/openmp-support_22.html](for almost a year now).  While Mac does not appear to support it in the default toolchain, I've been using Homebrew's `clang-omp` just fine.

I think that the BLIS build system should test for OpenMP support "the old fashioned way" (i.e. like `configure` does) and use OpenMP if it is available.

devinamatthews:
Options to configure have been added for:
- Setting the internal BLIS and BLAS/CBLAS integer sizes.
- Enabling and disabling the BLAS and CBLAS layers.

Additionally, configure options which require defining macros (the above plus the threading model), write their macros to the automatically-generated bli_config.h file in the top-level build directory. The old bli_config.h files in the config dirs were removed, and any kernel-related macros (SIMD size and alignment etc.) were moved to bli_kernel.h. The Makefiles were also modified to find the new bli_config.h file.

Lastly, support for OMP in clang has been added (closes #56).


fgvanzee:
Thanks for these improvements, Devin.

PR #70, opened 2016-04-25
Give the level1v operations some love

devinamatthews:
- Add missing axpby and xpby operations (plus test cases).
- Add special case for scal2v with alpha=1.
- Add restrict qualifiers.
- Add special-case algorithms for incx=incy=1.


fgvanzee:
@devinamatthews Before I merge, I would like to talk through some things first. I'll look for you tomorrow.

jeffhammond:
I fully support BLIS assuming C99 `restrict` is always available, but in the event you want to be conservative, you can `#define BLI_RESTRICT restrict`...

PR #74, opened 2016-05-10
Default-initialize all extern global variables to avoid generating common symbols.

### LINKED ISSUE #73: Constants in bli_const.c "undefined" on OSX
It seems that the linker on OSX is not able to find any of the constants defined in `bli_const.c`, such as BLIS_ONE etc. The problem, according to [this page](http://stackoverflow.com/questions/19398742/os-x-linker-unable-to-find-symbols-from-a-c-file-which-only-contains-variables) is that undefined constants end up as "common" symbols which are ignored by the OSX linker. The two available solutions seem to be:

1. Initialize the variables. Since these are `obj_t`'s, default-initialization with `...={}` should be OK (and they are initialized for real later).
2. Compile with `-fno-common`. This would have to be added to each configuration or in `common.mk`.

I am not sure when this problem first appeared, since I thought I had successfully compiled after the "big commit", but I am seeing it in my branch based off of cbcd0b739dc54bd14fbb46aeda267c26725cd70f. Using gcc-5.3.0 from Homebrew instead of the system "gcc" may also be a contributing factor.

devinamatthews:
Fixes #73. The list of candidates was compiled with `nm -o lib/libblis.a 2>/dev/null | grep "C _"`.

### DIFF
--- a/frame/1/packv/bli_packv_cntl.c
+++ b/frame/1/packv/bli_packv_cntl.c
@@ -34,7 +34,7 @@
 
 #include "blis.h"
 
-packv_t* packv_cntl;
+packv_t* packv_cntl = NULL;
 
 void bli_packv_cntl_init( void )
 {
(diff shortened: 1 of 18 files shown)


fgvanzee:
Thanks for this patch, Devin.

PR #81, opened 2016-05-30
use GCC intrinsic instead of pthread_mutex

jeffhammond:
I'm not sure how necessary this is, but I was reading the thread barrier code and couldn't help but implement this.

I can port this to use C11 atomics as well with trivial effort.

### DIFF
--- a/frame/base/bli_threading_pthreads.c
+++ b/frame/base/bli_threading_pthreads.c
@@ -44,14 +44,18 @@ int pthread_barrier_init(pthread_barrier_t *barrier, const pthread_barrierattr_t
     barrier->n_threads = count;
     barrier->sense = 0;
     barrier->threads_arrived = 0;
+#ifdef BLIS_USE_PTHREAD_MUTEX
     pthread_mutex_init( &barrier->mutex, NULL );
+#endif
     return 0;
 }
 
 int pthread_barrier_destroy(pthread_barrier_t *barrier)
 {
     if( barrier == NULL ) return 0;
+#ifdef BLIS_USE_PTHREAD_MUTEX
     pthread_mutex_destroy( &barrier->mutex );
+#endif
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


tlrmchlsmth:
@jeffhammond Which compilers support this? Just gcc? 
@devinamatthews Can we detect if this stuff is supported at configure time and turn this feature on if it is?

devinamatthews:
@jeffhammond I have a C++11 version of [mutex](https://github.com/devinamatthews/tblis/blob/master/src/core/tblis_mutex.hpp) and [barrier](https://github.com/devinamatthews/tblis/blob/master/src/core/tblis_thread.hpp) in TBLIS -- these should be quite similar to the C11 version.

@tlrmchlsmth `__sync_add_and_fetch` seems to be supported by gcc, clang, icc, and xlc. Otherwise, we could add optional macros in `bli_kernel.h` or maybe compiler-specific settings in `common.mk` to define alternatives if the gcc extensions are not supported. I highly doubt @fgvanzee will want to require C11 support.

jeffhammond:
At least, GCC, Clang, Intel, IBM and Cray support this intrinsic. Given
this code path is only used on Mac, only the first three really matter.

The first four support the full interface. Cray didn't have some of the GCC
intrinsics but they may have been fixed. If you find my OpenPA fork on
Github, you can see in the Cray branch the functions that need workarounds.

It is very easy to ifdef for C11, and Cray might have full C11 atomics
support. They use C11 in their libfabric implementation...

PGI is the compiler I know the least about, because I have no reason to use
it.



-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

devinamatthews:
Note that on OSX there is also OSSpinLock which has progressive back-off 
and no context-switch for short waits.

jeffhammond:
Indeed, using Mac-specific functions may be better than POSIX. But is the
extra maintenance effort worth the performance on small-core-count laptops?
I guess it's worth testing. But since Mac doesn't support affinity, they
are already losing performance there.

-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

devinamatthews:
I would vote for avoiding pthread_mutex altogether in favor of atomics, 
but I just wanted to offer an easy, ready-made option.

PR #82, opened 2016-05-30
Add thread yield function for spin-loops.

devinamatthews:
I noticed that TBLIS outperforms BLIS by about 30 GFLOP/s (400 vs. 370) on 12 cores of Lonestar 5 (both using OMP). To try and fix this I added a `pause` instruction in the spin-loop for Intel architectures, but alas there is no improvement. In any case, the optional macro `BLIS_YIELD` that I added in `bli_kernel.h` might be useful.

@fgvanzee, @tlrmchlsmth we should go over the respective codes at some point and figure out why TBLIS is faster.

### DIFF
--- a/config/dunnington/bli_kernel.h
+++ b/config/dunnington/bli_kernel.h
@@ -36,6 +36,9 @@
 #define BLIS_KERNEL_H
 
 
+#define BLIS_YIELD bli_intel_yield
+
+
 // -- LEVEL-3 MICRO-KERNEL CONSTANTS -------------------------------------------
 
 #undef  BLIS_SIMD_ALIGN_SIZE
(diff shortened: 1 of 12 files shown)


jeffhammond:
Is the 8% deviation reproducible within a single job, i.e. the same node over a short time interval?

I'll see if I can reproduce locally.

gaming-hacker:
what happened to this?  why wasn't merged?

devinamatthews:
@gaming-hacker I haven't measured any performance improvement from `pause` (which isn't too surprising since it is best not to use HyperThreads with BLIS), so there isn't a real rush to merge. Also, it would probably behoove me to add a similar feature for ARM at the same time.

jeffhammond:
FYI the OpenBLAS have been looking at the same issue, but don't have a clear conclusion: https://github.com/xianyi/OpenBLAS/pull/1051

PR #85, opened 2016-07-22
Change -openmp to -fopenmp for icc.

devinamatthews:
Since Intel 15, the `-openmp` option is depracated in favor of `-qopenmp`, and recent versions complain loudly about it. However, `-fopenmp` was and is still supported on Linux and OS X, providing coverage back to Intel 12.1. If at some point Intel nixes that too, then we can always switch again to `-qopenmp`.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -170,8 +170,8 @@ ifeq ($(THREADING_MODEL),auto)
 THREADING_MODEL := omp
 endif
 ifeq ($(THREADING_MODEL),omp)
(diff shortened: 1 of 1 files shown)


jeffhammond:
ICC supports a lot of GCC flags for compatibility reasons and I don't
foresee that changing.

-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

PR #86, opened 2016-07-22
Remove alignment restrictions on C in haswell kernel.

devinamatthews:
1) Change `vmovaps` to `vmovups` in C update.
2) Remove alignment checks. Check for row-storage is just `cmpq $8,%rsi` now.
3) For `beta != 0`, fuse load of C with FMA.

According to Agner Fog and other sources, `vmovups` is the preferred method for unaligned access (instead of `vinsertf128` etc.) for Ivy Bridge and later. This change is similar to part of the changes in #36, but without the Sandy Bridge vs. Ivy Bridge problem.


fgvanzee:
@devinamatthews I was just looking at the haswell micro-kernel code for unrelated reasons. Isn't it the case that *all* of the `vmovaps`/`vmovapd` be converted to `vmovups`/`vmovupd`? Or is there a reason we should keep the aligned instructions around for certain cases?

devinamatthews:
There is no harm in keeping them for (full vector) loads of A and/or B, and one could argue that using `vmovap[sd]` for that case exposes possible bugs in the alignment of the packing buffers. But from my understanding the performance is the same for Haswell and later.

fgvanzee:
@devinamatthews Sorry, I should have been more specific: I was referring only to the output use cases.

I see that we left `vmovaps` in the "general stored, beta zero" output cases. Do you see any reason these shouldn't be converted to unaligned instructions?

devinamatthews:
How can general storage use packed stores?

devinamatthews:
If you mean the register-to-register moves then vmovapd is fine.

fgvanzee:
Oh, obviously. Someone wasn't paying attention that an output macro followed. Please disregard. :)

Issue #75, opened 2016-05-10
configure does not obviously fail if non-existent configuration is used


devinamatthews:
For example, if one does `configure knc` instead of `configure mic`, the results is:

```
[dam879@stampede knc]$~/src/blis/configure -p `pwd`/install knc
configure: checking whether we need to update the version file.
configure: checking version file '/home1/02742/dam879/src/blis/version'.
configure: starting configuration of BLIS 0.2.0.
configure: manual configuration requested.
configure: configuring with 'knc' configuration sub-directory.
configure: using install prefix '/home1/02742/dam879/build/blis/knc/install'.
configure: debug symbols disabled.
configure: disabling verbose make output, enable with 'make V=1'.
configure: building BLIS as a static library.
configure: threading is disabled.
configure: the CBLAS compatibility layer is disabled.
configure: the BLAS compatibility layer is enabled.
configure: the internal integer size is automatically determined.
configure: the BLAS/CBLAS interface integer size is 32-bit.
configure: creating ./config.mk from /home1/02742/dam879/src/blis/build/config.mk.in
configure: creating ./bli_config.h from /home1/02742/dam879/src/blis/build/bli_config.h.in
configure: creating ./obj/knc
configure: creating ./obj/knc/config
configure: creating ./obj/knc/frame
configure: creating ./obj/knc/testsuite
configure: creating ./lib/knc
configure: mirroring /home1/02742/dam879/src/blis/config/knc to ./obj/knc/config
ls: cannot access /home1/02742/dam879/src/blis/config/knc: No such file or directory
configure: mirroring /home1/02742/dam879/src/blis/frame to ./obj/knc/frame
configure: creating makefile fragment in /home1/02742/dam879/src/blis/config/knc
ls: cannot access /home1/02742/dam879/src/blis/config/knc: No such file or directory
ls: cannot access /home1/02742/dam879/src/blis/config/knc: No such file or directory
/home1/02742/dam879/src/blis/build/gen-make-frags/gen-make-frag.sh: line 230: /home1/02742/dam879/src/blis/config/knc/.fragment.mk: No such file or directory
ls: cannot access /home1/02742/dam879/src/blis/config/knc: No such file or directory
configure: creating makefile fragment in /home1/02742/dam879/src/blis/frame
configure: creating makefile fragment in /home1/02742/dam879/src/blis/frame/0
configure: creating makefile fragment in /home1/02742/dam879/src/blis/frame/0/copysc
[...]

PR #93, opened 2016-09-17
Adds sanity check to configuration choice.

ShadenSmith:
This ensures that the chosen configuration is valid. In reference to #75.

### DIFF
--- a/configure
+++ b/configure
@@ -424,6 +424,14 @@ main()
 
 			echo "${script_name}: manual configuration requested."
 			config_name=$1
+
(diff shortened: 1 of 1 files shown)


ShadenSmith:
Are there any issues with this pull request?

fgvanzee:
Yes, thank you Shaden. This simply fell through the cracks. Thanks for the reminder.

ShadenSmith:
No problem, I figured as much. Thanks for looking over it.

PR #105, opened 2016-10-25
Support for Intel Knight's Landing

devinamatthews:

PR #106, opened 2016-10-25
KNL libm

jeffhammond:
This is trivial.  It depends on https://github.com/flame/blis/pull/105.

Hopefully @devinamatthews doesn't have any objections.


devinamatthews:
Can you do me a solid and check this for the other Intel configs too?

jeffhammond:
@devinamatthews Well, this change is only going to affect the KNL build but I guess you are asking me to do it holistically and submit a new PR.  Is that correct?  That's a perfectly reasonable suggestion.

jeffhammond:
I'm going to close this and create a new one that is (1) free of merge commits and (2) works for all Intel targets.

PR #107, opened 2016-10-26
never use libm with Intel compilers

jeffhammond:
Intel compilers include a highly optimized math library (libimf) that
should be used instead of GNU libm.

yes, this change is for ALL targets, including those that are not
supported by the Intel compiler.  there is no harm in doing this, and it
is future-proof in the event that the Intel compilers support other
architectures.

this has not been tested.  it should not be merged until that happens.

### DIFF
--- a/config/armv7a/make_defs.mk
+++ b/config/armv7a/make_defs.mk
@@ -77,7 +77,9 @@ ARFLAGS        := cru
 # --- Determine the linker and related flags ---
 LINKER         := $(CC)
 SOFLAGS        := -shared
+ifneq ($(CC_VENDOR),icc)
 LDFLAGS        := -lm
+endif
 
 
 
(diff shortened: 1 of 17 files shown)


jeffhammond:
Verified with `icc` for `auto reference dunnington sandybridge haswell`...

jeffhammond:
@devinamatthews You okay with this?

devinamatthews:
Some of the checks aren't strictly necessary (I'm totally gonna use icc on POWER7...), but are harmless where not applicable. This checking could also be done globally in common.mk. But I say merge.

jeffhammond:
@fgvanzee Do you want to pull the merge lever or should I do it?

fgvanzee:
Thanks Jeff. Merging now.

Issue #101, opened 2016-10-21
Threading should be tested in Travis CI.


devinamatthews:
It seems that commit 970745a introduced a bug which only manifests when threading is enabled (see #100). Build configurations for openmp and pthreads threading should be added to Travis CI to catch these kinds of bugs.

jeffhammond:
And it's easy to test many versions of GCC in Travis CI Mac instances using Homebrew.  Obviously, one can install RPMs in Linux instances with sudo enabled, but that's not ideal for other reasons.

fgvanzee:
Once, at a conference in Boston I think, Jeff taught me me the phrase, "We look forward to your patch." :)

I'm open to this, but I see it as a relatively low priority given how infrequently I dork with the multithreading functionality, and how I am usually quite thorough about testing those changes. (Granted, 970745a was an exception.)

jeffhammond:
@fgvanzee You may look forward to my patch 😉 

I don't always compute, but when I do, it's in parallel 😆

devinamatthews:
Fixed by #108.

Issue #26, opened 2015-05-21
Undefined references to ``GOMP_parallel``


poulson:
When I try to test for BLIS's support for dgemm_, I see the following link error, which does not seem to be covered by your build system wiki:

"""
/usr/bin/gcc-4.9   -DCHECK_FUNCTION_EXISTS=dgemm_    CMakeFiles/cmTryCompileExec3961259244.dir/CheckFunctionExists.c.o  -o cmTryCompileExec3961259244 -rdynamic /home/poulson/Install/lib/libblis.a -lpthread -lm
/home/poulson/Install/lib/libblis.a(bli_init.o): In function `bli_init':
bli_init.c:(.text+0xb4): undefined reference to`GOMP_critical_name_start'
bli_init.c:(.text+0xc6): undefined reference to `GOMP_critical_name_end'
/home/poulson/Install/lib/libblis.a(bli_init.o): In function`bli_finalize':
bli_init.c:(.text+0x1d0): undefined reference to `GOMP_critical_name_start'
bli_init.c:(.text+0x1e2): undefined reference to`GOMP_critical_name_end'
/home/poulson/Install/lib/libblis.a(bli_mem.o): In function `bli_mem_acquire_m':
bli_mem.c:(.text+0x77): undefined reference to`GOMP_critical_name_start'
bli_mem.c:(.text+0x8f): undefined reference to `GOMP_critical_name_end'
/home/poulson/Install/lib/libblis.a(bli_mem.o): In function`bli_mem_release':
bli_mem.c:(.text+0xfa): undefined reference to `GOMP_critical_name_start'
bli_mem.c:(.text+0x112): undefined reference to`GOMP_critical_name_end'
/home/poulson/Install/lib/libblis.a(bli_threading_omp.o): In function `bli_level3_thread_decorator._omp_fn.0':
bli_threading_omp.c:(.text+0x5): undefined reference to`omp_get_thread_num'
/home/poulson/Install/lib/libblis.a(bli_threading_omp.o): In function `bli_level3_thread_decorator':
bli_threading_omp.c:(.text+0x99): undefined reference to`GOMP_parallel'
collect2: error: ld returned 1 exit status
"""

fgvanzee:
Are you linking with `-fopenmp`?

poulson:
Nope, should I? (https://code.google.com/p/blis/wiki/BuildSystem#Linking_to_BLIS)

fgvanzee:
[...]

PR #108, opened 2016-11-01
Update .travis.yml with additional tests

### LINKED ISSUE #27: Sandybridge configuration incorrect for recent Mac OS X
Because of the fact that Apple chose to alias `gcc` to `clang` and `g++` to `clang++`, the current Sandybridge configuration, https://github.com/flame/blis/blob/ec25807b26da943868f0d0517c3720e50181b8f9/config/sandybridge/make_defs.mk#L79, will fail since Clang does not (yet) support OpenMP.

devinamatthews:
- Test knl configuration (without running of course).
- Test openmp and pthreads threading for auto configuration with 4 threads.
- Test auto configuration with and without pthreads on OSX.
- Also, run make in parallel.

I don't know how the `addons:` section works on OSX; hopefully it is just ignored. Also I should note that in theory this fixes #27 and #101 (and maybe #26?), and replaces PR #28.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ os:
 
 compiler:
   - gcc
+  - clang
 
 env:
   - RUN_TEST=1 THREADING="none" BUILD_CONFIG="auto"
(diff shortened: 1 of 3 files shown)


devinamatthews:
KNL build is still failing but Travis is helpfully not giving me a log...

jeffhammond:
I assume it is missing `hbwmalloc.h` if you didn't install MEMKIND.

jeffhammond:
https://travis-ci.org/flame/blis/jobs/172385632 shows 
```
./config/knl/bli_kernel.h:46:23: fatal error: hbwmalloc.h: No such file or directory
```

jeffhammond:
Instead of installing MEMKIND, just grab the header you need...
```
wget https://github.com/memkind/memkind/blob/dev/include/hbwmalloc.h
```

devinamatthews:
Ah, I see Travis is sporadically deciding to show logs again

devinamatthews:
I will just mark it as allowed to fail for now, and make memkind a soft dependency in the future.

jeffhammond:
Ok.  I will fix the KNL memkind issue once you merge this PR.

devinamatthews:
OK, there seem to be two things up with the OSX builds:

1. The `reference` configuration won't accept clang as the compiler.
2. The configuration auto-detection doesn't seem to be working since it is trying to use `reference`.

I can fix 1) fairly easily but I'm not sure what's up with 2).

devinamatthews:
Looks like everything is working now. @fgvanzee now all it needs is your blessing.

PR #109, opened 2016-11-04
Add automatic loop thread assignment.

devinamatthews:
- Number of threads is determined by BLIS_NUM_THREADS or OMP_NUM_THREADS, but can be overridden by BLIS_XX_NT as before.
- Threads are assigned to loops (ic, jc, ir, and jc) automatically by weighted partitioning and heuristics, both of which are tunable via bli_kernel.h.
- All level-3 BLAS covered.


devinamatthews:
@fgvanzee any thoughts on this?

PR #113, opened 2017-01-27
Change default threading parameters for KNL.

devinamatthews:
The default default parameters are tuned for Haswell, add better parameters for KNL.


devinamatthews:
Should really have been part of #109; I will just go ahead and merge since it is harmless.

PR #121, opened 2017-05-03
allow KNL build without hbwmalloc (i.e. emulated)

jeffhammond:
we want to be able to run BLIS KNL binaries on non-KNL machines via SDE.
although it is possible to install hbwmalloc implementation on such
systems, it is easier not to, since obviously the performance of SDE
execution is not representative so there is no reason to emulate HBW
allocation.

### DIFF
--- a/config/knl/bli_kernel.h
+++ b/config/knl/bli_kernel.h
@@ -43,10 +43,21 @@
 #define BLIS_SIMD_SIZE                   64
 #define BLIS_SIMD_NUM_REGISTERS          32
 
+#ifdef BLIS_NO_HBWMALLOC
+
+#include <stdlib.h>
+
+#define BLIS_MALLOC_POOL malloc
+#define BLIS_FREE_POOL free
+
+#else
+
 #include <hbwmalloc.h>
 
 #define BLIS_MALLOC_POOL hbw_malloc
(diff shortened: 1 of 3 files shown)


devinamatthews:
If we were using autoconf we could check for memkind in configure like sane people.

jeffhammond:
Neither the macro name nor the `-d sde` option name were optimal, but I couldn't think of anything better.  I figure that @fgvanzee's OCD will kick in at some point and he will find optimal names.

While I am a big fan of Autotools, the current build system is elegant in its simplicity.  Given that `configure` is a shell script, it can certainly try to find `hbwmalloc.h` and `libmemkind.*`.  Perhaps we just tell the user to set `MEMKIND_PATH` or both `MEMKIND_CPPFLAGS` and `MEMKIND_LDFLAGS` when targeting KNL, if they want these to be active.

I can produce another PR for this in a bit.

devinamatthews:
The autoconf comment was largely tongue-in-cheek, but one could compare 
for example to how memkind is detected in TBLIS.

fgvanzee:
@jeffhammond Thanks, Jeff. I'm currently working on changes to the build system to accommodate multi-architecture builds, and so I am appreciating that simplicity, too.

As for the cpp macro name, I'm fine with that.

Issue #122, opened 2017-05-03
Test all Intel CPUs in Travis CI via SDE


jeffhammond:
@devinamatthews You will probably like this...

Intel [SDE](http://www.intel.com/software/sde) supports no less than the following list of architectures:
```
     -quark              Set chip-check and CPUID for Intel(R) Quark
     -p4                 Set chip-check and CPUID for Pentium4
     -p4p                Set chip-check and CPUID for Pentium4 Prescott
     -mrm                Set chip-check and CPUID for Merom
     -pnr                Set chip-check and CPUID for Penryn
     -nhm                Set chip-check and CPUID for Nehalem
     -wsm                Set chip-check and CPUID for Westmere
     -snb                Set chip-check and CPUID for Sandy Bridge
     -ivb                Set chip-check and CPUID for Ivy Bridge
     -hsw                Set chip-check and CPUID for Haswell
     -bdw                Set chip-check and CPUID for Broadwell
     -skx                Set chip-check and CPUID for Skylake Server
     -skl                Set chip-check and CPUID for Skylake Client
     -cnl                Set chip-check and CPUID for Cannonlake
     -knl                Set chip-check and CPUID for Knights Landing
     -slt                Set chip-check and CPUID for Saltwell
     -slm                Set chip-check and CPUID for Silvermont
     -glm                Set chip-check and CPUID for Goldmont
```

This means that one can run native binaries for these processors on other processors, e.g. whatever Travis CI happens to be using.

I use it on a daily basis to emulate KNL on HSW.  It will allow us to set `RUN_TEST=1` for SNB, HSW, KNL, etc.

I am working on https://github.com/jeffhammond/blis/tree/travis-ci-sde-knl-testing.

devinamatthews:
Yes, I've used SDE for testing KNL and that other one. My concern would be build time of course as it is quite slow. Maybe only testing a subset of operations?

jeffhammond:
@devinamatthews Compilation is unaffected, because cross-compiler works fine.  We should probably have an option to test only 100x100 but I was running KNL binaries on HSW yesterday and it seemed pretty fast.  I'll time it today.  The only unknown is whether something SDE uses is perturbed by the AWS environment.  But I can time SDE inside of a Debian VM locally if necessary.

devinamatthews:
I tend to get in the O(0.1 GFLOP) range, which takes a while with all 
the tests that are run by default. But the proof is in the pudding.

jeffhammond:
[...]

PR #124, opened 2017-05-04
Travis CI SDE testing

jeffhammond:
This pull request implements https://github.com/flame/blis/issues/122.

This overhauls most of the Travis CI stuff.  In particular:
- I moved most of the interesting stuff into https://github.com/jeffhammond/blis/blob/travis-ci-sde-knl-testing/build/travis_driver.sh because it makes it a lot easier to script sensibly there.
- I determine what the native architecture is (https://github.com/jeffhammond/blis/blob/travis-ci-sde-knl-testing/build/travis_driver.sh#L18) and run tests natively if possible.  _This may be redundant with `BUILD_CONFIG=auto`._
- I run the tests in SDE if native execution isn't possible but emulation is (https://github.com/jeffhammond/blis/blob/travis-ci-sde-knl-testing/build/travis_driver.sh#L65).

As noted by @devinamatthews, performance with SDE is ~0.02 GF/s, so we probably need to prune the tests if we want them to finish in under an hour.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,18 +13,18 @@ env:
   - RUN_TEST=1 THREADING="none" BUILD_CONFIG="auto"
   - RUN_TEST=1 THREADING="none" BUILD_CONFIG="reference"
   - RUN_TEST=1 THREADING="none" BUILD_CONFIG="dunnington"
-  - RUN_TEST=0 THREADING="none" BUILD_CONFIG="sandybridge"
-  - RUN_TEST=0 THREADING="none" BUILD_CONFIG="knl"
-  - RUN_TEST=0 THREADING="none" BUILD_CONFIG="haswell"
+  - RUN_TEST=1 THREADING="none" BUILD_CONFIG="sandybridge"
+  - RUN_TEST=1 THREADING="none" BUILD_CONFIG="knl"
+  - RUN_TEST=1 THREADING="none" BUILD_CONFIG="haswell"
   - RUN_TEST=0 THREADING="none" BUILD_CONFIG="bulldozer"
   - RUN_TEST=0 THREADING="none" BUILD_CONFIG="piledriver"
   - RUN_TEST=0 THREADING="none" BUILD_CONFIG="carrizo"
   - RUN_TEST=0 THREADING="openmp" BUILD_CONFIG="auto"
   - RUN_TEST=0 THREADING="pthreads" BUILD_CONFIG="auto"
-  
+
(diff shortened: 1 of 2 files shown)


devinamatthews:
Also note that the travis_driver.sh script checks /proc/cpuinfo which won't work on OSX.

jeffhammond:
I properly guarded access to `/proc/cpuinfo` in https://github.com/jeffhammond/blis/blob/travis-ci-sde-knl-testing/.travis.yml#L66.  I forgot to add the guard in https://github.com/jeffhammond/blis/blob/travis-ci-sde-knl-testing/build/travis_driver.sh because Travis CI only tests `auto` on Mac and thus those statements will only be executed on Linux.

I will fix this.

devinamatthews:
How about these test input files?

[Archive.zip](https://github.com/flame/blis/files/976768/Archive.zip)

jeffhammond:
@devinamatthews Thanks.  I'll incorporate them shortly.

devinamatthews:
@jeffhammond can you elaborate on `${SDE_LOCATION}`?

jeffhammond:
`SDE_LOCATION` is a link to my personal Dropbox.  SDE requires one to accept license terms to use, and there is no way to enforce that with a direct download, so it was determined that a private link to e.g. Dropbox was the most effective way to get SDE into the Travis environment.

There are likely better methods.  I'm happy to accept suggestions.

jeffhammond:
Now that I think of it, I can make the link public if I password-protect the file and then hide the password in a Travis secure env var, but I have a bad experience with that in the context of doing that with Intel compiler license keys when trying to support Intel compilers in Travis.

devinamatthews:
This has been implemented in the `travis-virtualization branch` and is almost fully working now. As it turns out, there is a little wrinkle in emulating the AMD architectures because glibc and libm actually do their own CPUID detection and dispatch, and end up executing FMA4 instructions. Believe it or not, the only way to work around this is to apply a binary patch to ld.so!

jeffhammond:
What about using MUSL instead of GLIBC?  MUSL may not have all the CPU-specific optimizations and therefore behave better.

devinamatthews:
I've got a workaround in place, now I just have to deal with the kernel security flags so that sde will run.

devinamatthews:
Closing in favor of #212, which incorporates all of the contributions here. Thanks @jeffhammond for the idea and the crucial details.

PR #128, opened 2017-05-17
Portability and clang

iotamudelta:
A few trivial changes to the make system. Replace /bin/bash shebang with /usr/bin/env bash (which is portable to systems that have bash installed in, e.g., /usr/local/bin/bash). Mark a bunch of AMD architectures compilable with clang. Please note, that clang throws warnings for some unsupported optimization flags.

With these changes, BLIS compiles and passes testsuite on FreeBSD 12-CURRENT w/ carrizo config, BLAS and CBLAS, and pthread thread model.

Independent of this, it would be very nice if BLIS could be compiled into a single libblis.a for all applicable architectures as that would make it much easier to package.


jeffhammond:
I recommend that you add

> Independent of this, it would be very nice if BLIS could be compiled into a single libblis.a for all applicable architectures as that would make it much easier to package.

as an issue.  This is of interest to me too and I believe someone may already be working on it.

fgvanzee:
@iotamudelta Thank you for your contributions.

The feature you requested (the one that Jeff commented on) is one that I am actively working on. I call it runtime kernel management, which will allow supporting multiple architectures within one library file. I'll add it as an issue so that others know I'm working on it.

PR #130, opened 2017-05-24
Add optimized packing kernels.

devinamatthews:
Added SSE2 kernels for dunnington and AVX kernels for sandybridge/haswell. New kernels are enable for all of the configurations and have all been tested.

### DIFF
--- a/config/dunnington/bli_kernel.h
+++ b/config/dunnington/bli_kernel.h
@@ -112,6 +112,17 @@
 
 
 
+// -- LEVEL-1M KERNEL DEFINITIONS ----------------------------------------------
+
+// -- packm --
+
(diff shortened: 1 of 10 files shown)


devinamatthews:
OK, so it's only good for something like 1%, but oh well.

jeffhammond:
1% in DGEMM is not insignificant...

devinamatthews:
Apparently that 1% is more like +/- 1%...  also tried SW prefetches and that has a more or less random effect on performance as well. I'm going to close this until we understand the ramifications better.

fgvanzee:
Thanks for looking into this, Devin. Your experience is consistent with what I seem to remember from my (albeit dated) experiments with optimized packing. (That is, optimizing packing is harder than it looks, unfortunately.)

PR #138, opened 2017-07-07
Set missing free_fp in bli_membrk_init for free-ing GEN_USE buffers

hominhquan:
The membrk's free_fp is called when releasing GEN_USE buffers, but this free_fp is
not set in bli_membrk_init


fgvanzee:
@hominhquan Thank you for catching this and submitting this pull request.

PR #154, opened 2017-11-17
[WIP] Add x86 and x86_64 processor families.

iotamudelta:
As a suggestion: would such two meta-families be OK? And would the x86_64 correctly fall back to reference if no fitting specialized kernel is found for a given architecture?

### DIFF
--- a/config_registry
+++ b/config_registry
@@ -6,11 +6,10 @@
 #
 
 # Processor families.
-intel64:     haswell sandybridge penryn
-amd64:       zen excavator steamroller piledriver bulldozer
-arm32:       cortexa9 cortexa15
-x86:         reference
-x86_64:      haswell sandybridge penryn zen excavator steamroller piledriver bulldozer reference
+intel64:     haswell sandybridge penryn generic
+amd64:       zen excavator steamroller piledriver bulldozer generic
+arm32:       cortexa9 cortexa15 generic
+x86_64:      haswell sandybridge penryn zen excavator steamroller piledriver bulldozer generic
 
 # Intel architectures.
 haswell:     haswell
--- a/config_registry
+++ b/config_registry
@@ -9,6 +9,8 @@
 intel64:     haswell sandybridge penryn
 amd64:       zen excavator steamroller piledriver bulldozer
 arm32:       cortexa9 cortexa15
+x86:         reference
+x86_64:      haswell sandybridge penryn zen excavator steamroller piledriver bulldozer reference
 
 # Intel architectures.
 haswell:     haswell


fgvanzee:
@iotamudelta Thanks for this suggestion.

I have a few comments:
- I'm fine with having an `x86_64` configuration family. However, I don't think we need to define an `x86` configuration family given that BLIS does not currently support any microarchitectures (sub-configurations) that would belong in that family. So, let's nix that one.
- There is no sub-configuration named `reference` anymore; you probably meant to write `generic` instead. (Please see "Step 1" of the rewritten BuildSystem [1] wiki for more details. You may also be interested in the rewritten ConfigurationHowTo [2] wiki for a comprehensive nuts-and-bolts dive into configurations.)
- The `generic` configuration (the modern analogue of `reference` in previous versions) should not typically be included in a configuration family such as `x86_64`. Reference kernels are now built on a per-configuration basis, meaning _every_ sub-configuration gets its own set of reference kernels (partially because each sub-configuration might call for different compilation flags). However, easy access to some of the reference kernels is cut off when some of the context's kernel pointers are overwritten by pointers to optimized kernels, depending on the definition of `bli_cntx_init_<configname>()`. I will define a new function to more easily initialize a reference context for the active sub-configuration. But since each sub-configuration has its own set of compilation flags, very few people will ever need the `generic` sub-configuration if a microarchitecturally appropriate sub-configuration is present.

If you make these changes, I'll merge the pull request. Again, thanks for your involvement.

[1] https://github.com/flame/blis/wiki/BuildSystem
[2] https://github.com/flame/blis/wiki/ConfigurationHowTo

fgvanzee:
@iotamudelta I forgot to answer your second question. Basically, you are asking: If one configures BLIS for some configuration family but then links/runs the library on a system that does not correspond to any of the sub-configurations included in the family, what happens?

Let me think about / dig into this and get back to you.

iotamudelta:
@fgvanzee thanks for the comments!
* yes, x86 can go as a family.
* generic vs reference: didn't realize that, sorry!
* yes, that's my question. Application being to compile blis once for, say, x86_64 on a machine to include all possible kernels and then using those binaries on machines having both supported x86_64 architectures but maybe also unsupported ones (and correctly falling back to "generic" in that case). Use case is FreeBSD ports system where we want to distribute one library that works everywhere on x86_64 (and optimized for the supported architectures).

I'll update tonight with the changes. Should I hold out on including "generic" for now on x86_64 pending your investigation?

fgvanzee:
@iotamudelta I understand now. That makes sense re: your use case for the `generic` fall back functionality.

Yes, you can hold off on your changes until I figure out how BLIS should behave in the "unsupported sub-configuration via multi-configuration" build scenario.

jeffhammond:
>
>
>
>    - I'm fine with having an x86_64 configuration family. However, I
>    don't think we need to define an x86 configuration family given that
>    BLIS does not currently support any microarchitectures (sub-configurations)
>    that would belong in that family. So, let's nix that one.
>
>
Users can run 32-bit OS on newer hardware. If you want support in i686
Debian for example, you want such a config.
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

fgvanzee:
@iotamudelta It turns out your initial intuition was correct. 

After BLIS is configured with an Intel/AMD/x86_64 configuration family, `bli_cpuid_query_id()` is called at runtime (via `bli_arch_query_id()`) to choose the correct sub-configuration ID value. Even if the hardware does not test positive for any of the enabled sub-configurations, the function must of course still return an `arch_t` value. Thus, the only thing that would make sense to do would be to return a backstop value such as `BLIS_ARCH_GENERIC`. And if this is the behavior of `bli_cpuid_query_id()`, then configuration families must either implicitly or explicitly include the `generic` sub-configuration as one of its members. And based on the build system's new treatment of configurations, I think it makes more sense to keep that inclusion explicit and out in the open so that people see what's going on.

Thus, the `generic` sub-configuration should be included in every configuration family that uses `CPUID`--namely, `intel64`, `amd64`, and your proposed `x86_64`.

Thanks again for bringing this to my attention.

fgvanzee:
@jeffhammond Would the "32-bit OS on new hardware" scenario change the kernels that would be needed? Suppose someone was running a 32-bit Linux distribution on haswell hardware. Does that mean that we would have to fall back to kernels that don't use %rax, %rbx, etc? (I feel dumb for not intuitively knowing the answer to this already.)

jeffhammond:
> @jeffhammond <https://github.com/jeffhammond> Would the "32-bit OS on new
> hardware" scenario change the kernels that would be needed? Suppose someone
> was running a 32-bit Linux distribution on haswell hardware. Does that mean
> that we would have to fall back to kernels that don't use %rax, %rbx, etc?
> (I feel dumb for not intuitively knowing the answer to this already.)
>

I don’t know. Do they ever assume pointers or related types (eg size_t) are
64b?

-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

fgvanzee:
@jeffhammond According to [1], the GNU assembly suffix (e.g. `q` in `movq`) encodes the amount of data being loaded. Of course, this suffix is missing in Intel syntax, and so the size of the data is implied by the register name (64-bit `%rax` vs. 32-bit `%eax`). Given that a 32-bit Linux distribution would be using the 32-bit ABI, I'm guessing a micro-kernel that relied upon 64-bit integers and pointers would be problematic.

Bottom line: BLIS does not currently support 32-bit Intel/AMD/x86 systems.

[1] https://www.quora.com/What-is-the-difference-between-movq-and-movl-assembly-instruction

jeffhammond:
@fgvanzee Works for me.  I've been running 64-bit operating systems for ~10 years now.  I guess we can wait to see if anyone in the Linux package manager universe complains.

fgvanzee:
@jeffhammond Yeah, let's do that. I actually have one or two legacy 32-bit micro-kernels somewhere, but given that I no longer have 32-bit hardware, I have no way of properly testing any further development on that line of code.

iotamudelta:
@fgvanzee can I ask you to clarify the "BLIS does not support 32-bit"? I thought "generic" took care of that? That'd be an issue to make it default BLAS for FreeBSD.

iotamudelta:
@fgvanzee Updated the branch based on your input.

fgvanzee:
@iotamudelta Let me rephrase my comment that BLIS does not support 32-bit x86 systems: BLIS does not currently offer any sub-configurations that offer optimized kernels targeting 32-bit x86 hardware.

The `generic` sub-configuration can be tailored to support just about anything because it makes so few assumptions. As currently configured, `generic` is set up to output 64-bit object code via the `-m64` flag, but perhaps I should remove it so that it could also be used to provide support for 32-bit environments.

iotamudelta:
@fgvanzee yes, would be good if there was support for 32bit systems. We at FreeBSD (unfortunately, IMHO) still support them as tier 1 architectures. Either way, thanks for merging this and thanks for making BLIS more attractive for distributions!

devinamatthews:
@fgvanzee what does openblas do for 32-bit?

fgvanzee:
@devinamatthews According to the `QuickInstall.txt` file (actually, `GotoBLAS_02QuickInstall.txt`) of version 0.2.20, it inspects the binary type of gcc. If it's 64-bit, then it builds a 64-bit library. Otherwise it builds a 32-bit library. That's assuming the `make` variable `BINARY` was not explicitly set, as it is when you invoke the build system via either of the `quickbuild.??bit` scripts, or if you manually edit the `Makefile.rule` file before running `make`.

devinamatthews:
Mostly I mean does it use an optimited kernel and can it do AVX in 32 bit?

fgvanzee:
@devinamatthews Can it build AVX kernels on a 32-bit OS? Or can it build AVX kernels into 32-bit object code?

devinamatthews:
Is the latter case important? Both I guess.

fgvanzee:
From the `Makefile.system` of OpenBLAS 0.2.20:
```
# Force fallbacks for 32bit

ifeq ($(BINARY), 32)
ifeq ($(TARGET), HASWELL)
GETARCH_FLAGS := -DFORCE_NEHALEM
endif
ifeq ($(TARGET), SANDYBRIDGE)
GETARCH_FLAGS := -DFORCE_NEHALEM
endif
ifeq ($(TARGET), BULLDOZER)
GETARCH_FLAGS := -DFORCE_BARCELONA
endif
ifeq ($(TARGET), PILEDRIVER)
GETARCH_FLAGS := -DFORCE_BARCELONA
endif
ifeq ($(TARGET), STEAMROLLER)
GETARCH_FLAGS := -DFORCE_BARCELONA
endif
ifeq ($(TARGET), EXCAVATOR)
GETARCH_FLAGS := -DFORCE_BARCELONA
endif
ifeq ($(TARGET), ZEN)
GETARCH_FLAGS := -DFORCE_BARCELONA
endif
endif
```
So, it would appear that if `BINARY` is set to 32, such as when the user manually requests it via `Makefile.rule`, the build system forces an older microarchitectural target. Assuming that the maintainers of OpenBLAS aren't just doing this for fun, this would imply that AVX kernels cannot be built into 32-bit object code. It's not clear if the same applies to 32-bit OSes, but I'm guessing the answer there is "no" as well.

I tried to dig further, but their build system is difficult for me to follow.

devinamatthews:
So it looks like x86 should just use the `generic` config and anyone still using 32-bit will just get what they deserve.

iotamudelta:
@devinamatthews fair enough

@fgvanzee it seems the configure script is not working for generic targets:

configure: 'x86_64' is defined as having the following sub-configurations:
configure:    haswell sandybridge penryn zen excavator steamroller piledriver bulldozer generic
configure:    haswell sandybridge penryn piledriver bulldozer generic 
configure:   'haswell' is registered...and exists.
configure:   'sandybridge' is registered...and exists.
configure:   'penryn' is registered...and exists.
configure:   'zen' is registered...and exists.
configure:   'excavator' is registered...and exists.
configure:   'steamroller' is registered...and exists.
configure:   'piledriver' is registered...and exists.
configure:   'bulldozer' is registered...and exists.
configure:   'generic' is registered...and exists.
configure: checking sub-configurations' requisite kernels:
configure:   'haswell' kernels...exist.
configure:   'sandybridge' kernels...exist.
configure:   'penryn' kernels...exist.
configure:   'piledriver' kernels...exist.
configure:   'bulldozer' kernels...exist.
configure:   'generic' kernels...do NOT exist!
configure: 
configure: *** Cannot continue with nonexistent kernel 'generic'. ***
configure:  

looks like it tries to find kernels/generic and does not find that. Do you want to move ref_kernels there or change the logic in configure?

fgvanzee:
> looks like it tries to find kernels/generic and does not find that. Do you want to move
> ref_kernels there or change the logic in configure?

No. The `ref_kernels` directory must remain at the top level directory for nuanced reasons relating to the way object filepaths are generated and managed in the `Makefile`. 

I thought I fixed this issue already, but I'll look into it. Thanks for the feedback.

fgvanzee:
Okay, first data point is that I can't reproduce the error:
```
configure: manual configuration requested; configuring with 'x86_64'.
configure: checking configuration against contents of 'config_registry'.
configure: configuration 'x86_64' is registered.
configure: 'x86_64' is defined as having the following sub-configurations:
configure:    haswell sandybridge penryn zen excavator steamroller piledriver bulldozer generic
configure:    haswell sandybridge penryn piledriver bulldozer generic 
configure:   'haswell' is registered...and exists.
configure:   'sandybridge' is registered...and exists.
configure:   'penryn' is registered...and exists.
configure:   'zen' is registered...and exists.
configure:   'excavator' is registered...and exists.
configure:   'steamroller' is registered...and exists.
configure:   'piledriver' is registered...and exists.
configure:   'bulldozer' is registered...and exists.
configure:   'generic' is registered...and exists.
configure: checking sub-configurations' requisite kernels:
configure:   'haswell' kernels...exist.
configure:   'sandybridge' kernels...exist.
configure:   'penryn' kernels...exist.
configure:   'piledriver' kernels...exist.
configure:   'bulldozer' kernels...exist.
configure:   'generic' kernels...exist.
```
And actually, the problem is with line 935 of `configure`:
```
       if [ ! -d "${kernels_dirpath}/${kernel}" ]; then
```
The idea is that the error only prints if the kernel directory is not a directory (doesn't exist). But here's the problem: `git` does not track empty directories. So the easiest fix is to put a dummy file in there so that `git` will recognize `kernels/generic`.

fgvanzee:
@iotamudelta BTW, would you like to disclose your name so I can acknowledge you in my `git` commit log entries? Alternatively, I could just refer to you by your github handle. Either way is fine.

iotamudelta:
@fgvanzee oh, sure: Johannes M Dieterich. Met Devin at a workshop in CA and he made a passionate case for flame/blis.

fgvanzee:
@iotamudelta Thanks Johannes. Give 25eee3cc a try. Hopefully that was the only thing preventing line 935 from failing in your case. :)

fgvanzee:
@iotamudelta Oh, right. I forgot to ask for this in your initial pull request. :) I'll put it together.

fgvanzee:
@iotamudelta Try 6c3ba502. I've added the `x86_64` directory and tweaked some of the vectorization flags for some of the other Intel-related configurations.

iotamudelta:
@fgvanzee that compiles. I will need to do some work on the FreeBSD port before it'll install. Then I'll test and let you know. Looks good, though! Unrelated: I will need to investigate why powerpc64 on FreeBSD doesn't compile blis but that's probably just compiler stuff (we use gcc on that platform, IIRC).

fgvanzee:
@iotamudelta Re: powerpc64. Feel free to adjust the `power7` configuration to suite your needs, if you find that useful. I think you would be the only person using it.

iotamudelta:
@fgvanzee I am not even sure we do power7. Certainly power8 and I am aware of porting to power9 going on. Before making blis default for the entire portstree, there is a *lot* of architectures I need to get the generic configuration to work reliably (the specialized kernels I don't even care for that much right now, we are currently on reference Fortran BLAS/LAPACK). https://www.freebsd.org/platforms/ for a list of platforms.

iotamudelta:
@fgvanzee there is a new issue now, apparently it cannot setup the correct excavator kernel for my A12-8800B (and doesn't fall back to generic either):

libblis: frame/base/bli_gks.c (line 257):
libblis: Invalid architecture id value.
libblis: Aborting.

fgvanzee:
@iotamudelta Please open a new issue, and remember to include information on how you configured BLIS and what you were doing to trigger the error.

PR #181, opened 2018-03-23
(Example) Add vectorization over n to Zen sdotxf kernel.

devinamatthews:
Performance improvement is ~3x for the specific case targeted vs. something like 5x measured for vectorization over the other direction.

### DIFF
--- a/kernels/zen/1f/bli_dotxf_zen_int_8.c
+++ b/kernels/zen/1f/bli_dotxf_zen_int_8.c
@@ -94,7 +94,6 @@ void bli_sdotxf_zen_int_8
 	// operation as a loop over dotxv.
 	if ( b_n != fuse_fac )
 	{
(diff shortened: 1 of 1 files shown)


devinamatthews:
I just mean that GFLOPs(after) ~= 3*GFLOPS(before) when the new code path is exercised (n == 8 && lda == 1). I measured about 5x for the existing vectorized code path (these are on Tyler's i7 machine). This makes sense because I have to do two loads per iteration instead of 1.125 on average.

fgvanzee:
I'm still having trouble following along. Furthermore, the performance I observe from the new `dotxf` code is almost identical to the old code. Let's table this until we're both in the office.

fgvanzee:
Thanks for this work, Devin. I've applied your redesign to both of the zen dotxf kernels (`sdotxf` and `ddotxf`).

PR #198, opened 2018-05-06
Some cross compilations fixes

kali:
* allow user to specify ranlib executable
* gcc and clang version can be reliably obtain from -dump-version (version string parsing is broken by: `arm-unknown-linux-gnueabi-gcc (crosstool-NG crosstool-ng-1.23.0-240-gaec73d9c-dirty) 7.2.0`)


configure
```diff
@@ -1039,7 +1040,12 @@ get_compiler_version()
 	# The last part ({ read first rest ; echo $first ; }) is a workaround
 	# to OS X's egrep only returning the first match.
 	cc_vendor=$(echo "${vendor_string}" | egrep -o 'icc|gcc|clang|emcc|pnacl|IBM' | { read first rest ; echo $first ; })
-	cc_version=$(echo "${vendor_string}" | egrep -o '[0-9]+\.[0-9]+\.?[0-9]*' | { read first rest ; echo ${first} ; })
+    if [ "$cc_vendor" == "gcc" -o "$cc_version" == "clang" ]
```
devinamatthews: `$cc_version` -> `$cc_vendor`. You could also add icc since it supports `-dumpversion` too.

devinamatthews: Also, only `=` is POSIX sh, not `==`.

devinamatthews:
@fgvanzee? Looks great to me.

fgvanzee:
@kali Thanks for these contributions.

One last request: Can you revert the update to `version`? (Our policy is to only update `version` during version bump commits.)

kali:
yeah, I can't say i really understand what updated it actually.

fgvanzee:
@kali It wasn't you; the build system automatically updates `version` when `configure` is run so that the file includes a "patch" number (e.g. the "9" in 0.3.2-9). It's just that the file is not meant to track the "patch" number from commit to commit.

Thanks for the quick tweak. I'll merge now.

PR #206, opened 2018-05-14
Debian config

nschloe:
Include the Debian packaging config. This way, anyone can find it back.


fgvanzee:
Thanks @nschloe . Is it okay if I (eventually) relocate this to a packages subdirectory?

nschloe:
For a proper Debian package, the `debian` folder must be in the root directory of the package. The main purpose of having it in this repo is that other people can easily find it back, so yeah, relocating it won't be a problem.

fgvanzee:
@nschloe If it really should stay at the root level of the distribution, we can leave it. I was simply trying to declutter in advance of other packages being included.

Especially since there are no other directories that begin with `d` (tab completion), we can leave it at the root level and revisit when we get other packages. :)

PR #209, opened 2018-05-15
Fix shared library builds on platforms other than Linux and macOS

ararslan:
The way the logic is currently laid out, any platform that isn't Linux gets assigned the .dylib shared library extension and the macOS-specific compiler flags. This reverses the logic to check for macOS first, and have the fallback use the Linux definitions, which apply to most other systems as well.


fgvanzee:
Looks good to me. Thanks for catching this issue, Alex.

PR #212, opened 2018-05-26
Test all x86_64 configurations*...

devinamatthews:
...using Intel SDE. Some things to note:

- gcc-6 and binutils-2.26 must be loaded (and the system ld and as replaced with newer versions) so that no configurations are blacklisted.
- `sudo: required` is needed so that SDE will run.
- Dropbox is used to host SDE with a secret API token (don't echo it!).
- Because glibc/libm use cpuid to do their own runtime detection, care must be taken to disable this for the AMD configs, otherwise FMA4 instructions are executed which SDE doesn't do. A binary patch is applied to the system ld.so, libc.so, and libm.so.

* except bulldozer

### DIFF
--- a/travis/do_testsuite.sh
+++ b/travis/do_testsuite.sh
@@ -8,8 +8,8 @@ export BLIS_JC_NT=1
 export BLIS_IR_NT=1
 export BLIS_JR_NT=1
 
-make BLIS_ENABLE_TEST_OUTPUT=yes testblis
+make testblis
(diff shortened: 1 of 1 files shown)


devinamatthews:
@fgvanzee if you'll take a look to familiarize yourself with the changes then I'll merge to `dev`.

fgvanzee:
Thanks @devinamatthews and @jeffhammond for all your work on this. (Travis CI is being ornery right now and putting out false test fails, but we're confident that it will pass once the service is working again.)

PR #216, opened 2018-06-05
Make get_b_next return b_next instead of a_next

tlrmchlsmth:


fgvanzee:
Thanks for catching this, Tyler.

PR #231, opened 2018-07-06
Disable SDE for PRs

devinamatthews:
Pull requests cannot use Travis secret variables, so SDE needs to be disabled. This PR should suffice as a test.


devinamatthews:
Seems to work.

fgvanzee:
Thanks Devin!

PR #242, opened 2018-08-29
fix compilation of armv7a kernels

kali:
Hello there,

This fixes two small compilation issues in armv7a intrinsic kernels.


fgvanzee:
Thanks for this, Mathieu!

PR #244, opened 2018-09-06
add an adhoc impl for pthread_barrier

kali:
Fix for #243

### DIFF
--- a/testsuite/src/test_libblis.h
+++ b/testsuite/src/test_libblis.h
@@ -54,7 +54,7 @@
 
 // For pthreads API.
 #include <pthread.h>
-#ifdef __APPLE__
+#if !defined(_POSIX_BARRIERS) || (_POSIX_BARRIERS < 0)
 #include "pthread_barrier.h"
 #endif
 
--- a/testsuite/src/test_libblis.h
+++ b/testsuite/src/test_libblis.h
@@ -54,6 +54,9 @@
 
 // For pthreads API.
 #include <pthread.h>
+#ifdef __APPLE__
+#include "pthread_barrier.h"
+#endif
 
 //
 // --- Constants and types -----------------------------------------------------
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
   # macOS with system compiler (clang)
   - os: osx
     compiler: clang
-    env: OOT=0 TEST=0 SDE=0 THR="none" CONF="auto"
+    env: OOT=0 TEST=1 SDE=0 THR="none" CONF="auto"
 install:
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo rm -f /usr/bin/as; fi
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo ln -s /usr/lib/binutils-2.26/bin/as /usr/bin/as; fi
--- a/testsuite/src/pthread_barrier.h
+++ b/testsuite/src/pthread_barrier.h
@@ -1,0 +1,68 @@
+#ifdef __APPLE__
+
+#ifndef PTHREAD_BARRIER_H_
+#define PTHREAD_BARRIER_H_
+
+#include <pthread.h>
+#include <errno.h>
+
+typedef int pthread_barrierattr_t;
+typedef struct
+{
+    pthread_mutex_t mutex;
+    pthread_cond_t cond;
+    int count;
+    int tripCount;
+} pthread_barrier_t;
+
+
+inline int pthread_barrier_init(pthread_barrier_t *barrier, const pthread_barrierattr_t *attr, unsigned int count)
+{
+    if(count == 0)
+    {
+        errno = EINVAL;
+        return -1;
+    }
+    if(pthread_mutex_init(&barrier->mutex, 0) < 0)
+    {
+        return -1;
+    }
+    if(pthread_cond_init(&barrier->cond, 0) < 0)
+    {
+        pthread_mutex_destroy(&barrier->mutex);
+        return -1;
+    }
+    barrier->tripCount = count;
+    barrier->count = 0;
+
+    return 0;
+}
+
+inline int pthread_barrier_destroy(pthread_barrier_t *barrier)
+{
+    pthread_cond_destroy(&barrier->cond);
+    pthread_mutex_destroy(&barrier->mutex);
+    return 0;
+}
+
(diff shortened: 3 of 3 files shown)


devinamatthews:
`pthread_barrier_t` has been there forever, but it is an optional POSIX feature, none of which are implemented by Apple.

fgvanzee:
> `pthread_barrier_t` has been there forever, but it is an optional POSIX feature, none of which are implemented by Apple.

Haha, because of course it is optional.

Thanks Devin.

devinamatthews:
@kali do the current Travis CI tests catch this (pre-fix)? If not could you please update them?

devinamatthews:
@fgvanzee it is because Apple hates interoperable standards and only implements enough of POSIX to get UNIX certification...

fgvanzee:
> @fgvanzee it is because Apple hates interoperable standards and only implements enough of POSIX to get UNIX certification...

So obnoxious.

kali:
@devinamatthews tests were not built and not ran (TEST=0) on macos, so the problem were not appearing. I was able to reproduce it locally on my laptop in order to fix it (with the latest xcode version).

The second commit re-enables travis tests on macos, I'm confident switching the two commits would make the problem appear on travis.

devinamatthews:
@kali OK, I think that will cover it. @fgvanzee ready to accept?

isuruf:
Thanks for fixing the issue.

testsuite/src/test_libblis.h
```diff
@@ -54,6 +54,9 @@
 
 // For pthreads API.
 #include <pthread.h>
+#ifdef __APPLE__
```
isuruf: How about checking for `_POSIX_BARRIERS`? See http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/unistd.h.html

fgvanzee: I like this idea better than branching on `__APPLE__`, even if the two are functionally equivalent for now. @devinamatthews ?

devinamatthews: `_POSIX_BARRIERS`

fgvanzee: @kali Would you mind following Isuru's recommendation and changing your cpp guard to one that uses `_POSIX_BARRIERS`?

Given that the document states, "This symbol shall always be set to the value 200809L," I assume you can change line 57 above to
```c
#ifdef _POSIX_BARRIERS
```

kali: well, it's actually defined at -1 on macos...

fgvanzee: @kali BTW, thanks for pointing out the -1 behavior. It wasn't clear from the header webpage that -1 was a possible value given that the description claimed it was always set to 200809L.

fgvanzee: > @kali BTW, thanks for pointing out the -1 behavior. It wasn't clear from the header webpage that -1 was a possible value given that the description claimed it was always set to 200809L.

This is probably Apple being rebellious wherever they can. Care to comment, @devinamatthews, since you have more experience with standards?

jeffhammond:
@fgvanzee If you think Pthread barrier is obnoxious, look up the Apple affinity API 😉

testsuite/src/test_libblis.h
jeffhammond: If you believe in the absolute truth of POSIX preprocessor symbols, then do the following:
```
#if defined(_POSIX_BARRIERS) && (_POSIX_BARRIERS == 200809L)
```
On the other hand, if you have ever experienced a platform that does not correctly define preprocessor symbol or has bugs in system libraries, then you you will write a configure test to determine that the implementation actually works, or at least compiles and links, and use a macro like `BLIS_HAS_PTHREAD_BARRIER`.

jeffhammond: At the risk of shaking anyone's faith in standards, see https://wolandscat.net/2018/06/22/standards-paper-as-toilet-paper-torvalds/.

fgvanzee: Thanks for your comments, Jeff.

I agree that the "right" way is to write a configure test to determine if it actually works. However, I don't have the patience for that right now, and anything I don't have the patience for I don't usually expect of others (though I am delighted when they go above and beyond). :)

jeffhammond: If it helps to understand my cynicism 

From page 40 of the Red Book "IBM System Blue Gene Solution: Blue Gene/P Application Development":
> 4.4 Shared memory support
> 
> Shared memory is supported in Dual mode and Virtual Node Mode. Shared memory in SMP mode is not necessary because each processor already has access to all of the node’s memory. Shared memory is allocated using standard Linux methods (shm_open() and mmap()).

Translation: application that use POSIX shared-memory without taking this into account will crash because `shm_open` does a check for SMP mode and returns without doing what it is supposed to, and no preprocessor macro will tell you this.

I can't remember if the documentation above was added before or after I had to root cause this bug in a Quantum Monte Carlo code by reading the operating system source code for a day.

fgvanzee:
> @devinamatthews I don't claim to know Apple's motivations but as @jedbrown will likely tell you, Pthread barriers are a bad API and nobody should use them anyways, except when they are lazy and do not care about performance.

I totally agree that we basically don't need pthread's barriers. I did indeed use it only because I was lazy and didn't care about performance. (It was for the testsuite, to synchronize simulated application threads before moving from one operation to the next.) However, this whole discussion is making me second-guess even that isolated case of low-standards-itis.

> If you think Pthread barrier is obnoxious, look up the Apple affinity API

I'll remember that for the next time my blood pressure and stress levels are dangerously low.

fgvanzee:
@kali Thanks for that change. I'm ready to merge, but we first have to resolve a conflict in the `.travis.yml` file. (Can you see that "dialog box" in the conversation? Or maybe it's just people with write privileges.)

I think I know how to resolve it, but if you can do it, I'll let you do the honors since you committed the "offending" code in PR #245. :)

fgvanzee:
@jeffhammond It appears the Travis CI test for this wasn't finished when you merged (and later failed). Was it just one of those false alarms where Travis falls apart during the environment setup for no apparent reason?

devinamatthews:
@fgvanzee the git checkout failed because the PR had already been merged. False alarm.

fgvanzee:
@devinamatthews Good to know, thanks Devin.

jeffhammond:
All of the testing on master has passed since this merge.
itsfine

PR #245, opened 2018-09-06
cortexa9 and cortexa53 travis build + qemu test

kali:
Adds travis build and test for one arm32 target and one aarch64.

One sidenode: as ubuntu/trusty cross gcc is a bit ancient, i had to downgrade the mtune argument from a53.a57 to a57 only... I'm not sure if we should have two differenc configs just for that or if we can use a more recent ubuntu..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,12 +34,25 @@ matrix:
   - os: osx
     compiler: clang
     env: OOT=0 TEST=0 SDE=0 THR="none" CONF="auto"
+  # cortexa15 build and test (qemu)
+  - os: linux
+    compiler: arm-linux-gnueabihf-gcc
+    env: OOT=0 TEST=FAST SDE=0 THR="none" CONF="cortexa15" \
+      PACKAGES="gcc-arm-linux-gnueabihf qemu-system-arm qemu-user" \
+      TESTSUITE_WRAPPER="qemu-arm -cpu cortex-a15 -L /usr/arm-linux-gnueabihf/"
+  # cortexa57 build and test (qemu)
+  - os: linux
+    compiler: aarch64-linux-gnu-gcc
+    env: OOT=0 TEST=FAST SDE=0 THR="none" CONF="cortexa57" \
+      PACKAGES="gcc-aarch64-linux-gnu qemu-system-arm qemu-user" \
+      TESTSUITE_WRAPPER="qemu-aarch64 -L /usr/aarch64-linux-gnu/"
 install:
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo rm -f /usr/bin/as; fi
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo ln -s /usr/lib/binutils-2.26/bin/as /usr/bin/as; fi
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo rm -f /usr/bin/ld; fi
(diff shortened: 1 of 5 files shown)


devinamatthews:
Sweet! I knew something like this was possible but I never wanted to wade through the details to make it work. Thanks.

fgvanzee:
@kali Very nice. Thanks for these changes, Matthieu.

fgvanzee:
> One sidenode: as ubuntu/trusty cross gcc is a bit ancient, i had to downgrade the mtune argument from a53.a57 to a57 only... I'm not sure if we should have two differenc configs just for that or if we can use a more recent ubuntu..

I recently looked into this, and I seem to recall that trusty was the latest Ubuntu available on Travis CI (which I found very surprising). @devinamatthews, would you mind commenting on / confirming this?

devinamatthews:
Trusty is indeed the latest on Travis, but it might be possible to get a newer cross GCC from a ppa. Probably doesn't matter for testing though.

kali:
Well, I did change the kernel makefile itself to switch from a57.a53 to a53. This will impact the main build itself, not only the tests. Just wanted to make sure you folks were aware of that too.

devinamatthews:
Oh, I have no idea what effect that has.

fgvanzee:
Perhaps @figual could comment. Fran, what do you think of @kali's change?

kali:
Well maybe the right answer is actually to give a53 a separate configuration, with the same kernel (the header indicates it was benched on a53). a53 looks much more frequent in the wild than the a57.a53 combination. And a53 would be testable on travis...

fgvanzee:
@kali I actually like the idea of creating a separate sub-configuration. But let's clarify something first. Above, you said

> Well, I did change the kernel makefile itself to switch from a57.a53 to a53

But originally you said

> i had to downgrade the mtune argument from a53.a57 to a57

The PR code would seem to agree with neither: it appears you switched from a57.a53 to a57. :)

So that us non-ARM experts can follow along, can you reaffirm which one looks more frequent in the wild, and which one is testable on Travis?

kali:
all right sorry for the confusion :) I'm not an expert myself, but I'm trying to learn.

cortex-a53 is relatively common and cheap (raspberry pi 3, odroid c2, ...).
cortex-a57 is a bit more high-end (nvidia boards, nintendo switch, ...)

Both can be targetted with -mtune on gcc-4.8, on travis.

Now, a57.a53 is a big.LITTLE setup, where cores of two variants are co-existing on the same processor, allowing to switch from a low-consumption mode to a high performance mode. A few Android phones are using it, but Apple, as often, has done its own thing. This setup is not targettable on travis, and it was the one targetted by the config "cortexa57" before my change. I switched the config to targetting plain a57.  

Finally, the gemm micro kernels we have (referred to as armv8) look like they have been optimized with both a53 and a57 in mind, and benched on both, according to the comments.

fgvanzee:
@kali Thanks for that very helpful clarification. I'm fine switching the `cortex-a57` configuration to use flags that target only `a57`, assuming Francisco does not raise any concerns or objections.

As for creating a separate `cortex-a53` configuration, I'm still on board. In fact, I can get started on that while we wait for Fran to comment.

fgvanzee:
@kali I've added a `cortexa53` sub-configuration in the `dev` branch (fb81c7fc). It is identical to `cortexa57` except that it uses `-mtune=cortex-a53` and `-mcpu=cortex-a53` in its configuration's `make_defs.mk` file.

PR #256, opened 2018-09-27
configure: Replace '/' with '#' in sed expressions.

cdluminate:
This avoids the following error, which results in `make` failure.

```
configure: creating ./config.mk from ./build/config.mk.in
sed: -e expression #1, char 22: number option to `s' command may not be zero
configure: creating ./bli_config.h from ./build/bli_config.h.in
```


cdluminate:
Reference: https://stackoverflow.com/questions/43196365/bash-sed-e-expression-1-char-21-number-option-to-s-command-may-not-be-zer/43196398

what's actually happening there
```shell
+ echo 'configure: creating ./config.mk from ./build/config.mk.in'
configure: creating ./config.mk from ./build/config.mk.in
+ cat ./build/config.mk.in
+ sed -e s/@version@/upstream/0.4.1-4/g
+ sed -e s/@so_version_major@/1/g
+ sed -e s/@config_name@/x86_64/g
sed: -e expression #1, char 22: number option to `s' command may not be zero
+ sed -e 's/@config_list@/skx knl haswell sandybridge penryn zen excavator steamroller piledriver bulldozer generic/g'
+ sed -e 's/@kernel_list@/skx knl haswell sandybridge penryn zen piledriver bulldozer generic/g'
+ sed -e s/@CXX@/g++/g
+ sed -e s/@CC_VENDOR@/gcc/g
+ sed -e 's/@kconfig_map@/bulldozer:bulldozer generic:generic haswell:haswell knl:knl penryn:penryn piledriver:piledriver sandybridge:sandybridge skx:skx zen:zen/g'
+ sed -e s/@CC@/gcc/g
+ sed -e s/@ldflags_preset@//g
+ sed -e s/@RANLIB@/ranlib/g
+ sed -e s/@dist_path@/./g
+ sed -e s/@os_name@/Linux/g
+ sed -e s/@debug_type@/off/g
+ sed -e s/@configured_oot@/no/g
+ sed -e s/@enable_verbose@/no/g
+ sed -e 's/@install_incdir@/\/home\/lumin\/blis\/include/g'
+ sed -e s/@enable_memkind@/no/g
+ sed -e s/@enable_blas@/yes/g
+ sed -e 's/@install_libdir@/\/home\/lumin\/blis\/lib/g'
+ sed -e s/@enable_static@/yes/g
+ sed -e s/@enable_arg_max_hack@/no/g
+ sed -e s/@cflags_preset@//g
+ sed -e s/@threading_model@/no/g
+ sed -e s/@so_version_minorbuild@/0.0/g
+ sed -e s/@sandbox@//g
+ sed -e s/@enable_shared@/yes/g
+ sed -e s/@enable_cblas@/no/g
+ sed -e 's/@install_sharedir@/\/home\/lumin\/blis\/share/g'
+ set +x
```

fgvanzee:
@cdluminate Can you describe a bit about your build environment? I've never run into this issue before. We took measures to escape the slashes in all substituted text that might contain slashes (from file paths), hence the variables named with the `_esc` suffix. Hopefully you can help us better understand why those accommodations were not sufficient in your case.

cdluminate:
If you clone https://salsa.debian.org/science-team/blis and `./configure x86_64` it would fail to generate a correct `config.mk`. The source code is identical to `blis-0.4.1` except for the debian directory. After removing `.git` directory it starts to work again.

I didn't look into this problem but replacing the characters used in sed expression is the fastest way to make it work.

fgvanzee:
@cdluminate Thanks for those details. I would really like to reproduce this issue before moving forward.

I cloned https://salsa.debian.org/science-team/blis as you suggested and found that the `configure` script appears out-of-date. I tried to follow the `git log` entry to see why this might be, but the log doesn't go back very far. Can you help me understand the history of this clone?

fgvanzee:
@cdluminate I've been able to reproduce the issue with your clone (and not with an upstream clone). So I have something to go on now in my debugging. (It may still be interesting to learn more about the history of the clone.)

Another question: is it Debian policy to only fetch updates from upstream when there is a version tag? Or can/do you also pull in updates in between versions as well?

fgvanzee:
@cdluminate I've found the problem. You had created a tag `upstream/0.4.1` in your local repository. If the `.git` directory is present, it's assumed by `configure` that you're working with a `git` clone and that `git` is available. In those cases, the `git` tag is queried by `configure` to determine the current version number. That version string is one of the values substituted into config.mk via `sed`.

Nowadays, the version number is less important since it is no longer integrated into the final library name (as it once was), and so its use is limited to `configure` output and embedding within the library itself so that it can be queried at runtime.

I've pushed 7d96fc4 to the `master` branch, which fixes this issue while keeping the original formatting of the sed statements. Please confirm that this works for your purposes at your next convenience.

Thanks for your help in tracking this down.

cdluminate:
> Another question: is it Debian policy to only fetch updates from upstream when there is a version tag? Or can/do you also pull in updates in between versions as well?

The git repository layout, as you see, is following Debian's git-buildpackage workflow, which is the most common practice in the Debian community. However it's not a part of Debian Policy at all.

* `upstream` branch holds an identical copy of upstream source. (except for upstrean's `debian` directory)  
* `pristine-tar` branch holds upstream tarballs, which extract into the same content as that in `upstream` branch
* `master` branch holds `upstream` branch plus `debian` directory.

Everytime a new upstream release comes out, I will download upstream tarball and update `upstream` and `pristine-tar` branches accordingly. That means this workflow is based on upstream tarball releases instead of upstream git.

See [1] for details about debian science team's workflow recommendation. And thank you for the fix. I'll test the fix soon and close this PR.

[1] https://science-team.pages.debian.net/policy/#idm297

fgvanzee:
> Everytime a new upstream release comes out, I will download upstream tarball and update `upstream` and `pristine-tar` branches accordingly. That means this workflow is based on upstream tarball releases instead of upstream git.

That answers my question. This will probably motivate me to put out formal releases a little more often. :) Thanks for that explanation.

PR #269, opened 2018-10-11
Fix OMP nesting problem.

### LINKED ISSUE #267: OpenMP nested parallelism bug in BLIS
There appears to be an threading issue with BLIS compiled with OpenMP and run inside a parallel nested loop.

It crashes the Elk code with a segmentation fault.

I can't reproduce the seg. fault with a small example, but the following program never terminates on our AMD machine:

```
program test
implicit none
integer, parameter :: n=20
integer i
real(8), allocatable :: a(:,:),b(:,:),c(:,:)

!$OMP PARALLEL DEFAULT(SHARED) &
!$OMP PRIVATE(a,b,c)
!$OMP DO
do i=1,100
  allocate(a(n,n),b(n,n),c(n,n))
  a=1.d0
  b=1.d0
  call dgemm('N','N',n,n,n,1.d0,a,n,b,n,0.d0,c,n)
  deallocate(a,b,c)
end do
!$OMP END DO
!$OMP END PARALLEL

end program
```

The above code has to be compiled with
```
gfortran -fopenmp test.f90 lapack.a libblis.a
```

Both the number of OpenMP and BLIS threads have to be larger than 1:

```
export OMP_NUM_THREADS=2
export BLIS_NUM_THREADS=2
```

If either OMP_NUM_THREADS=1 or BLIS_NUM_THREADS=1 then the code runs fine.

If BLIS is compiled with pthreads:
```
./configure -t pthreads zen
```
then the code also runs fine.

The official AMD release (https://developer.amd.com/amd-cpu-libraries/blas-library/) was compiled with OpenMP and therefore also has the bug.

devinamatthews:
Detect when OpenMP uses fewer threads than requested and correct accordingly, so that we don't wait forever for nonexistent threads. Fixes #267.

### DIFF
--- a/frame/thread/bli_thrcomm_openmp.c
+++ b/frame/thread/bli_thrcomm_openmp.c
@@ -237,9 +237,20 @@ void bli_l3_thread_decorator
 		{
 			if ( id == 0 )
 			{
-				n_threads = n_threads_real;
-				bli_thrcomm_init( gl_comm, n_threads );
+	            if ( n_threads_real != 1 )
+	            {
+	                bli_print_msg( "A different number of threads was "
+	                               "created than was requested.",
+	                               __FILE__, __LINE__ );
+	                bli_abort();
+	            }
+
+				n_threads = 1;
+				bli_thrcomm_init( gl_comm, 1 );
+				bli_rntm_set_num_threads_only( 1, rntm );
+				bli_rntm_set_ways_only( 1, 1, 1, 1, 1, rntm );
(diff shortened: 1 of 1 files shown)


devinamatthews:
I believe this is ready to merge. With both `OMP_NESTED=TRUE` and `OMP_DYNAMIC=TRUE` it might be possible to trigger the error condition, so if that comes up we will need to do a more thorough fix.

PR #287, opened 2018-11-28
Fix configuration links

SuperFluffy:
I was looking at the Kernel guide/howto and noticed that the links to the configuration howto didn't work. This fixes it.


fgvanzee:
Thanks for your feedback, @SuperFluffy. Also, I like your handle. :)

PR #289, opened 2019-01-05
Initialize error messages at compile time

hominhquan:
- Assigning strings directly to the bli_error_string array, instead of snprintf() at execution-time.

- From the programming point of view, this simplifies code and further adding of error code. 

- From the binary point of view, I believe this does not change the executable size, or even reduce it by some KB (all constant strings in `.data` section). The `bli_error_string`array will be moved from the `.bss` section to the `.data` one. And those constant hard-coded strings, which have always been in the `.data` section, instead of getting "duplicated" to the `bli_error_string` array at execution-time (function `bli_error_init_msgs()`), now should be directly put to the right place at compile time.

- From the performance point of view, we save dozens of sprintf.

### DIFF
--- a/build/libblis-symbols.def
+++ b/build/libblis-symbols.def
@@ -976,9 +976,6 @@ bli_dzxpbym_md_unb_var1
 bli_error_checking_is_enabled
 bli_error_checking_level
 bli_error_checking_level_set
-bli_error_finalize
-bli_error_init
-bli_error_init_msgs
 bli_error_string_for_code
 bli_ffree_align
 bli_ffree_noalign
@@ -990,6 +987,7 @@ bli_find_area_trap_l
 bli_fmalloc_align
 bli_fmalloc_align_check
 bli_fmalloc_noalign
+bli_fmalloc_post_check
 bli_fprintm
 bli_fprintm_check
 bli_fprintm_ex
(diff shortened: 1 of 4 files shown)


jeffhammond:
I like this.

hominhquan:
FYI, Travis failed on this commit due to an infrastructure problem (`apt-get` command failed). Can someone relaunch this build (@fgvanzee ) ? I do not have the right for that.

fgvanzee:
No worries, @hominhquan. I've restarted the two failed jobs.

I will try review your commit sometime today. Thanks for your contribution.

fgvanzee:
```c
static char bli_error_string[BLIS_MAX_NUM_ERR_MSGS][BLIS_MAX_ERR_MSG_LENGTH] =
{
	[-BLIS_INVALID_ERROR_CHECKING_LEVEL]      = "Invalid error checking level.",
	[-BLIS_UNDEFINED_ERROR_CODE]              = "Undefined error code.",
	[-BLIS_NULL_POINTER]                      = "Encountered unexpected null pointer.",
...
```
Neat. I didn't realize this was valid C syntax for statically initializing select elements of an array.

And I appreciate that you preserved the newlines between groups of related error codes!

fgvanzee:
I'm going to make a few more cleanups that are facilitated by these changes, then I'll merge.

PR #311, opened 2019-05-10
add info about CXX in configure

jeffhammond:
BLIS now uses a C++ compiler, albeit in a limited way.  The `CXX` environment variables controls this but isn't documented by `print_usage()`.

I am not sure it actually matters, but if the user is settings `CC` explicitly, they probably want to set `CXX` to the same toolchain.


fgvanzee:
Thanks for this PR, Jeff. (And also your patience while I was on vacation.)

PR #334, opened 2019-08-30
Always use sqsumv to compute normfv.

### LINKED ISSUE #332: snrm2 overflow for large input floats
DSNRM2 in BLIS overflows when applied to a an array `[1e20]` while it doesn't overflow in the Netlib BLAS (or OpenBLAS).

I run into this while trying to pass the scipy test suite with BLIS in https://github.com/scipy/scipy/issues/10744, as mentionned in that issue,

> Netlib BLAS does indeed [apply a scale in SNRM2](http://www.netlib.org/lapack/explore-html/df/d28/group__single__blas__level1_ga35c2ec0e9bfdaa545320c2a134fcc471.html#ga35c2ec0e9bfdaa545320c2a134fcc471) to avoir the overflow, but I don't know it it's actually part of the BLAS spec.

There are also a number of other scipy test failures due to overflow I think in other functions (see the linked issue).

devinamatthews:
Fixes #332, but don't merge until @rth confirms the problem is MacOS and not something else.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -83,6 +83,7 @@ but many others have contributed code and feedback, including
   Benda Xu              @heroxbd
   Costas Yamin          @cosstas
   Chenhan Yu            @ChenhanYu          (The University of Texas at Austin)
+  Roman Yurchak         @rth                (Symerio)
   M. Zhou               @cdluminate
 
 BLIS's development was partially funded by grants from industry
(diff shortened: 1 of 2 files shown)


rth:
It was on Linux (Gentoo) unfortunately https://github.com/flame/blis/issues/332#issuecomment-526712938

PR #339, opened 2019-09-13
AOCL 2.0

kvaragan:
Contains optimizations for EPYC-2.
Added new configurations for EPYC - 1 & EPYC-2
./configure auto - will automatically select appropriate AMD architecture
,/configure zen - EPYC - 1
./configre zen2  - EPYC-2


config/zen/make_defs.mk
```diff
@@ -32,53 +33,22 @@
...
-endif
-
-# Flags specific to optimized kernels.
-CKOPTFLAGS     := $(COPTFLAGS)
 ifeq ($(CC_VENDOR),gcc)
-# gcc 6.0 (clang 4.0) or later:
-#CKVECFLAGS     := -mavx2 -mfpmath=sse -mfma -march=znver1
-# gcc 4.9 (clang 3.5) or later:
-# possibly add zen-specific instructions: -mclzero -madx -mrdseed -mmwaitx -msha -mxsavec -mxsaves -mclflushopt -mpopcnt
-CKVECFLAGS     := -mavx2 -mfpmath=sse -mfma -march=bdver4 -mno-fma4 -mno-tbm -mno-xop -mno-lwp
```
devinamatthews: The `-mno-tbm -mno-xop -mno-lwp` flags are important for testing on Travis, since it will actually be running under Intel SDE. I highly doubt that they will affect performance in any way.

ravi-svrc: tbm, xop, lwp, fma4 are extensions which were enabled by default in older AMD architectures like bulldozer but they have been deprecated with Zen architectures. so, we don't need to explicitly use -mno-tbm -mno-xop -mno-lwp flags to disable them.

devinamatthews: Ah, OK. They would only be needed if using bdver4 as a fake znver1 on older gcc versions then (the updated version would require gcc 6.0+ or clang 4.0+).

fgvanzee:
@kvaragan Can you resubmit this PR request as a merge onto `flame:amd` instead of `flame:master`? I prefer to use `flame:amd` as a staging area where I make further edits and cleanups prior to merging into `flame:master`. Thanks!

PR #340, opened 2019-09-17
Fix description for function bli_*pxby2v

ShmuelLevine:
Revise the description of this function to reflect that the sum of the scaled elements of each vector are saved into a 3rd vector, rather than overwriting one of the original vectors.


fgvanzee:
Thanks @ShmuelLevine!

Issue #299, opened 2019-02-14
arm64 family support from OpenBLAS?


loveshack:
In case no-one's spotted it and the method would be useful since arm64 family has been dropped: I'd forgotten to say that OpenBLAS has acquired DYNAMIC_ARCH support, as they call it on Linux in
https://github.com/xianyi/OpenBLAS/commit/d5aeff636f2d8ba99d1e5ed511c3770970f440af
I don't know how much you can rely on the kernel support it needs.

fgvanzee:
@loveshack Could you clarify the gist of your comment? I'm having trouble understanding what you are suggesting.

loveshack:
I thought you needed a cpuid-type mechanism to dispatch on arm64 micro-arches, and that shows how to do it.
Apologies if that's wrong.
-- 
Sent from a pocket telegraph machine; excuse the result.

fgvanzee:
Oh, right! I understand now. Sorry I didn't follow at first. Yes, someone should look at what they do to solve this problem. (Not sure if that will be me, but we know one place to look now.) Thanks for pointing this out.

loveshack:
I added /proc-based support for arm32 (which isn't in OpenBLAS) to the PR.
That completes support for families of the supported architectures.  (See also a comment on #335.)

devinamatthews:
I think the action is in #344 now, closing.

PR #344, opened 2019-09-24
Arm64 micro-architecture dispatch

loveshack:
This provides a cpuid-like mechanism for arm64, per #299.
After doing it, I realized it presumably has limited effect because it doesn't affect block sizes, though it does affect the ISA for ThunderX2, which I haven't been able to try yet.  (OpenBLAS does have different kernel configurations for some of the micro-arches, but I don't know if significantly different.)
There are also comments about arm32 implementation if anyone is interested.

### DIFF
--- a/frame/base/bli_cpuid.c
+++ b/frame/base/bli_cpuid.c
@@ -401,6 +401,10 @@ arch_t bli_cpuid_query_id( void )
 			if ( bli_cpuid_is_cortexa57( model, part, features ) )
 				return BLIS_ARCH_CORTEXA57;
 #endif
+#ifdef BLIS_CONFIG_CORTEXA53
+		if ( bli_cpuid_is_cortexa53( model, part, features ) )
+			return BLIS_ARCH_CORTEXA53;
+#endif
 			// If none of the other sub-configurations were detected, return
 			// the 'generic' arch_t id value.
 			return BLIS_ARCH_GENERIC;
@@ -437,12 +441,7 @@ bool_t bli_cpuid_is_thunderx2
        uint32_t features
      )
 {
-	// Check for expected CPU features.
-	const uint32_t expected = FEATURE_NEON;
-
-	if ( !bli_cpuid_has_features( features, expected ) ) return FALSE;
-
-	return TRUE;
+	return model == BLIS_ARCH_THUNDERX2;
 }
 
 bool_t bli_cpuid_is_cortexa57
@@ -452,12 +451,7 @@ bool_t bli_cpuid_is_cortexa57
        uint32_t features
      )
 {
-	// Check for expected CPU features.
-	const uint32_t expected = FEATURE_NEON;
-
-	if ( !bli_cpuid_has_features( features, expected ) ) return FALSE;
-
-	return TRUE;
+	return model == BLIS_ARCH_CORTEXA57;
 }
 
 bool_t bli_cpuid_is_cortexa53
@@ -467,12 +461,7 @@ bool_t bli_cpuid_is_cortexa53
        uint32_t features
      )
 {
-	// Check for expected CPU features.
-	const uint32_t expected = FEATURE_NEON;
-
-	if ( !bli_cpuid_has_features( features, expected ) ) return FALSE;
-
-	return TRUE;
+	return model == BLIS_ARCH_CORTEXA53;
 }
 
 bool_t bli_cpuid_is_cortexa15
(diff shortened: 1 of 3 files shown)


devinamatthews:
@fgvanzee we should revisit this.

fgvanzee:
@loveshack What is the status of this PR? Is it ready to go as far as you know?

loveshack:
> @loveshack What is the status of this PR? Is it ready to go as far as you know?

Probably not now.  It was all tested to the extent I had virtual
hardware, but needs merging and testing again.  It also needs an
addition for a64fx/sve, presumably, and other micro-arches, like
neoverse.  It's unclear to me to what extent aarch64 variants should be
configured individually.

devinamatthews:
@loveshack Leaving out some uarches for now is fine. Getting the base code in is what is important from this PR I think. I did a remerge recently but you should check my changes.

devinamatthews:
@loveshack is this mergeable? I know you had wanted to add some stuff like neoverse but I think getting those can be handled in smaller PRs later.

devinamatthews:
@fgvanzee can you give this PR a final look? I think everything should pass now, except maybe the ARMSVE QEMU test (I can check this out if it does fail).

devinamatthews:
?? @fgvanzee all the Travis builds seem to be getting canceled?

fgvanzee:
> ?? @fgvanzee all the Travis builds seem to be getting canceled?

Actually, the first one failed after all of the jobs were in progress and running for a while. Then the second one was cancelled.

But yeah, no idea.

frame/base/bli_cpuid.c
```diff
@@ -1023,7 +925,309 @@ int vpu_count( void )
...
+   to POWER, instead of reading /proc, like:
+
+   #include <sys/auxv.h>
+   #include <linux/auxvec.h>  // for AT_PLATFORM
+   char * platform = NULL;
+   platform = (char*) getauxval (AT_PLATFORM);
+
+   that just yields "v7l" on cortexa9.
+
+ */
```
xrq-phys, line 1230: I'm wondering if these comments are sort of duplicating with [L994](https://github.com/loveshack/blis/blob/de156bd/frame/base/bli_cpuid.c#L994-L1050).

Maybe it'd be also good to add a comment here mentioning it as a fallback method for ARMv7?

xrq-phys:
Additionally, may I propose the following changes so that `feature` is correctly set under the new detection method?
(Although the variable seems not used at all.)

```diff
diff --git a/frame/base/bli_cpuid.c b/frame/base/bli_cpuid.c
index 3626f9bb..d704d825 100644
--- a/frame/base/bli_cpuid.c
+++ b/frame/base/bli_cpuid.c
@@ -956,7 +956,10 @@ et al
 #include <sys/sysctl.h>
 #endif
 
-static uint32_t get_coretype()
+static uint32_t get_coretype
+     (
+       uint32_t* features
+     )
 {
        int implementer = 0x00, part = 0x000;
 
@@ -983,7 +986,10 @@ static uint32_t get_coretype()
        }
        
        bool has_sve = getauxval( AT_HWCAP ) & HWCAP_SVE;
+       if (has_sve)
+         *features |= FEATURE_SVE;
 #endif //__linux__
+       *features |= FEATURE_NEON;
 
 #ifdef __APPLE__
        // Better values could be obtained from sysctlbyname()
@@ -1135,8 +1141,7 @@ uint32_t bli_cpuid_query
      )
 {
        *model    = MODEL_ARMV8;
-       *features = 0;
-       *part     = get_coretype();
+       *part     = get_coretype(features);
 
        return VENDOR_ARM;
 }
```

devinamatthews:
> Additionally, may I propose the following changes so that `feature` is correctly set under the new detection method?
> (Although the variable seems not used at all.)

I can add that, but at least for now the features aren't used in selecting the configuration (that is actually happening before `get_coretype()` returns...). `has_sve` is used there, but it really only kicks in for emulated SVE (QEMU) or some "unknown" SVE chip. In principle, I suppose the code should also check the SVE vector length?

I'll make the above changes for now and save these other things for PRs later, I just want to finally get this code in to master.

frame/base/bli_cpuid.c
devinamatthews: I actually hadn't looked at the ARMv7 part. I will give it a once-over.

devinamatthews:
@fgvanzee can you give this a final blessing?

devinamatthews:
@fgvanzee pinging again, this should be ready to merge.

fgvanzee:
Yes, sorry for the delay. This looks fine to me.

(We may need to resolve conflicts since the PR is now a few weeks old.)

fgvanzee:
No conflicts after all. Yay. :fireworks:

PR #361, opened 2019-11-07
Add Haiku to the known OS list

korli:


fgvanzee:
Thanks @korli. I've never heard of Haiku, but more OS support is always better!

korli:
thanks!

PR #379, opened 2020-03-19
Fixed bug in amaxv which returned incorrect index on constant input.

matcross:
Hi,

A constant vector with increment 1 and more than 3 elements gets the wrong amaxv index computed because of an incorrect comparison operator when initializing i_max_l.

We at NAG noticed this was affecting a few of our LAPACK test programs.

I'm afraid I didn't look into an associated test case for the BLIS framework. I do have a Fortran test program, if that helps at all:

```
   Program p
     Double Precision, Allocatable :: x(:)
     x = [ (42.D0,i=1,4) ]
     If (idamax(size(x),x,1)/=1) Stop 'FAIL'
     Print *, 'ok'
   End Program
```


matcross:
Cancelled in light of further issues having been discovered. This here doesn't seem to be the right solution for the whole of the story. See also #380.

PR #395, opened 2020-04-26
Fix a typo in KernelsHowTo.md

YingboMa:
I believe it should be
```
  C11 := beta * C11 + alpha * A1 * B1
```


fgvanzee:
Thanks for catching this, @YingboMa!

PR #396, opened 2020-04-28
New kernel set for Arm SVE using assembly

docularxu:
Here adds two kernels for Arm SVE vector extensions.
1. a gemm  kernel for double at sizes 8x8.
2. a packm kernel for double at dimension 8xk.

To achive best performance, variable length agonostic programming
is not used. Vector length (VL) of 256 bits is mandated in both kernels.
Kernels to support other VLs can be added later.

"SVE is a vector extension for AArch64 execution mode for the A64
instruction set of the Armv8 architecture. Unlike other SIMD architectures,
SVE does not define the size of the vector registers, but constrains into
a range of possible values, from a minimum of 128 bits up to a maximum of
2048 in 128-bit wide units. Therefore, any CPU vendor can implement the
extension by choosing the vector register size that better suits the
workloads the CPU is targeting. Instructions are provided specifically
to query an implementation for its register size, to guarantee that
the applications can run on different implementations of the ISA without
the need to recompile the code."  [1]

[1] https://developer.arm.com/solutions/hpc/resources/hpc-white-papers/arm-scalable-vector-extensions-and-application-to-machine-learning

Signed-off-by: Guodong Xu <guodong.xu@linaro.org>

### DIFF
--- a/frame/include/bli_arch_config.h
+++ b/frame/include/bli_arch_config.h
@@ -265,6 +265,9 @@ CNTX_INIT_PROTS( generic )
 
 // -- ARM architectures --
 
+#ifdef BLIS_KERNELS_ARMSVE
+#include "bli_kernels_armsve.h"
+#endif
 #ifdef BLIS_KERNELS_ARMV8A
 #include "bli_kernels_armv8a.h"
 #endif
--- a/kernels/armsve/1m/bli_dpackm_armsve256_asm_8xk.c
+++ b/kernels/armsve/1m/bli_dpackm_armsve256_asm_8xk.c
@@ -1,0 +1,235 @@
+/*
+
+   BLIS
+   An object-based framework for developing high-performance BLAS-like
+   libraries.
+
+   Copyright (C) 2014, The University of Texas at Austin
+   Copyright (C) 2020, Linaro Limited
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions are
+   met:
+    - Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    - Redistributions in binary form must reproduce the above copyright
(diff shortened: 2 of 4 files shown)


fgvanzee:
@docularxu Sorry for the delayed response. Thank you for this contribution! I checked the new files and it appears that you found, read, and followed the section of the [ConfigurationHowTo](https://github.com/flame/blis/blob/master/docs/ConfigurationHowTo.md) regarding [adding new kernel sets](https://github.com/flame/blis/blob/master/docs/ConfigurationHowTo.md#adding-a-new-kernel-set). Thanks for making my review easy!

We don't have access to ARM SVE hardware, so we can't test out the code. Furthermore, we have limited experience on ARM, so please consider yourself nominated as the maintainer of these new kernels going forward. :) But feel free to ask us questions should the need arise.

jeffhammond:
@fgvanzee https://developer.arm.com/tools-and-software/server-and-hpc/compile/arm-instruction-emulator, but I completely understand why you would not want to mess with this.

fgvanzee:
> but I completely understand why you would not want to mess with this.

Yeah, unfortunately I don't have time to take on a new project like this. But it's good to know such resources are available for others and/or for my own use in the future.

rvdg:
Unfortunately, we have to prioritize work for which we are actually sponsored.  ARM architectures do not qualify at the moment due to that restriction.

docularxu:
I am using 'qemu-aarch64' [1] for verification of these two kernels. And I will keep optimizing them when I get hands-on real hardware. 

So far, the only public commercially available ARM-SVE-enabled hardware is Fujitsu's A64FX [2]. A64FX has 512-bit vector length, which is different with what I submitted. 

[1] https://www.linaro.org/blog/sve-in-qemu-linux-user/
[2] https://github.com/fujitsu/A64FX

PR #403, opened 2020-05-20
avoid loading twice in armv8a gemm kernel

docularxu:
This bug happens at a corner case, when k_iter == 0 and we jump to
CONSIDERKLEFT.

In current design, first row/col. of a and b are loaded twice.

The fix is to rearrange a and b (first row/col.) loading instructions.

Signed-off-by: Guodong Xu <guodong.xu@linaro.org>

### DIFF
--- a/kernels/armv8a/3/bli_gemm_armv8a_asm_d6x8.c
+++ b/kernels/armv8a/3/bli_gemm_armv8a_asm_d6x8.c
@@ -108,13 +108,6 @@ __asm__ volatile
 " add x25,x24,x10                            \n\t" //Load address Column 10 of C
 " add x26,x25,x10                            \n\t" //Load address Column 11 of C
 "                                            \n\t"
-" ldr q0, [x0]                               \n\t"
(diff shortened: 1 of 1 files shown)


docularxu:
To test the correctness, I tested dgemm and dtrmm for matrix sizes from 1 to 480. All tests passed.

fgvanzee:
@docularxu Thanks for this patch! I don't have much experience with ARM assembly, so I'll let @dnparikh @figual comment, if they would like to.

figual:
Looks good to me, good catch!!

fgvanzee:
@docularxu My PI, @rvdg, was wondering if you would be willing to share your latest performance data, especially if that data set compares BLIS versus ARMPL. I know how much time benchmarking can take, so if this is not something you can do at this time, just let us know.

docularxu:
@fgvanzee @rvdg ARMPL is commercially licensed, and I don't have license. That's one problem.

fgvanzee:
@docularxu No worries, then! Thanks again for your bugfix.

PR #422, opened 2020-07-13
New GEMM Assembly & Configuration Set for Arm SVE 

xrq-phys:
This pull request contains 2 parts:
- A new GEMM assembly for 512-bit vector length of Arm SVE;
- A new sub-configuration for Arm SVE that determines vector length at runtime with `incd` instruction and selects GEMM kernels accordingly.

The new PACKM kernel is not registered yet. Sorry for that.

As this pull request can be separated into 2 parts (with the sub-config depending on layout of kernels), I'm not sure if it was good to create pull request this way. If not, I'll close this and open another one with 512-bit assembly kernel only.

### DIFF
--- a/config/armsve/bli_cntx_init_armsve.c
+++ b/config/armsve/bli_cntx_init_armsve.c
@@ -1,0 +1,95 @@
+/*
+
+   BLIS
+   An object-based framework for developing high-performance BLAS-like
+   libraries.
+
+   Copyright (C) 2014, The University of Texas at Austin
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions are
+   met:
+    - Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    - Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    - Neither the name(s) of the copyright holder(s) nor the names of its
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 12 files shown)


xrq-phys:
Sorry but according to the latest benchmark this kernel & config is yet to be optimized.
I'm retracting this for the moment.

xrq-phys:
Dear BLIS developers,

Hello again.

I'm reopening this pull request with a few updates:
- A few improvements on the ARM SVE-512 kernel;
- A new `dgemm` kernel specialized for A64fx chip (reason below);
- A new configuration called `a64fx`;

**Reason for a different `dgemm` kernel for A64fx**

`dgemm_armsve512_asm` kernel under `kernels/armsve/3` is mainly composed of SVE indexed `FMLA` instructions (opcode=`0x64`). This strategy is the same as `dgemm_armsve256_asm` kernel located at the same directory. It is able to increase the interval between a vector's load and its reference by `FMLA`. However, actual profiling result of that kernel gives (*Test size: GEMM(2400,1600,500)*):

Bmk_DGEMM

Left part of combo histogram shows that in most time the processor is only committing 1 or 2 instructions while A64fx has 2 FP pipelines and 2 integer pipelines summing up to 4. This fact drastically lowers final GFlOps yielded (c.f. spread at the end). However, FP stall rate and memory/cache access wait is quite low, indicating no impediment to FP pipelines.

Though not documented in materials disclosed by Fujitsu, I suspect the FP pipeline in A64fx does not have element duplicator for indexed SVE operations so that one single `0x64 FMLA` is executed with both FP pipelines, each half-occupied. As a workaround, another kernel is created for A64fx with `0x64 FMLA` replaced with `0x65 FMLA` and it does yield higher GFlOps:

Bmk_DGEMM 2

| | BLIS + 0x64 Kernel | BLIS + 0x65 Kernel | Fujitsu BLAS |
| ---- | ---- | ---- | ---- |
| DGEMM(2400, 1600, 500) | 24 GFlOps | 33 GFlOps | 42 GFlOps |

PR #429, opened 2020-07-23
Add BLIS_EXPORT_BLIS to bli_abort.

### LINKED ISSUE #428: Quest: undefined reference to 'blis_abort'
hello，
        After installing the blis with configure 'zen', I tryied #include "blis.h" in my code. But, when I compile it with command "gcc -fopenmp -lm -lblis *.c -o a.out", I receive the error of ld as written by my title.
        Besides, when I tried commands :  "gcc -fopenmp -c *.c -o *.o" & "gcc *.o $(PREFIX)/lib/zen/libblis.a -fpenmp -lm -o a.out", I make it successfully. How to solve  or explain this question?

devinamatthews:
Fixes #428.

### DIFF
--- a/frame/base/bli_error.h
+++ b/frame/base/bli_error.h
@@ -39,7 +39,7 @@ BLIS_EXPORT_BLIS void     bli_error_checking_level_set( errlev_t new_level );
 BLIS_EXPORT_BLIS bool_t    bli_error_checking_is_enabled( void );
 
 void      bli_print_msg( char* str, char* file, guint_t line );
-void      bli_abort( void );
+BLIS_EXPORT_BLIS void      bli_abort( void );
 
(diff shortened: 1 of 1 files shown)


fgvanzee:
Thanks for identifying this fix, Devin. But I can't quite figure out what was leading it to be needed in the first place.

devinamatthews:
My understanding is that there is at least one inline function in blis.h that calls bli_abort. Since such a function would result in a call to bli_abort *in user code*, then the symbol must be exported in the .so.

fgvanzee:
> Since such a function would result in a call to bli_abort *in user code*, then the symbol must be exported in the .so.

This is the case even if the function is a static/inline function that is never invoked in the user code?

devinamatthews:
Without optimization, yes. This would break user debug builds.

PR #436, opened 2020-08-06
Add checks so that s390x is detected as 64-bit.

### LINKED ISSUE #423: failure to detect 64-bit systems
The s390x Fedora package wasn't getting built for the 64-bit target. The patch for blis 0.7 at https://src.fedoraproject.org/rpms/blis/raw/master/f/blis-s390x.patch addresses s390x specifically and tries to be more general, but it probably isn't robust.

devinamatthews:
Fixes #423.

### DIFF
--- a/frame/include/bli_system.h
+++ b/frame/include/bli_system.h
@@ -63,7 +63,7 @@
 
 // Determine if we are on a 64-bit or 32-bit architecture.
 #if defined(_M_X64) || defined(__x86_64) || defined(__aarch64__) || \
-    defined(_ARCH_PPC64)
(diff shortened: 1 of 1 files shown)


devinamatthews:
@fgvanzee can I get a thumbs-up? I could also merge little things like this directly if you don't mind.

fgvanzee:
Seems simple and unlikely to break anything. My favorite kinds of commits.

fgvanzee:
Yes, you are welcome to merge commits like this. Of course, if/when there is any doubt, just @ me.

PR #345, opened 2019-09-24
POWER micro-architecture dispatch

loveshack:
This implements a Linux-specific, compiler-neutral selection, unlike the (possibly?) gcc-specific simple one you didn't like. It currently only affects the compilation of generic code.

Note that there's a strange problem I've had to kludge round by lowering the optimization levels noted in a log message, but it seems worth recording this now.  I can only try with Fedora 28's gcc 8.3 (on power8) currently, but will try to get the system updated to a supported Fedora.  I might be able to get access to power9, but haven't tried.

This was branched from the arm64 equivalent as there originally was a merge clash with bits I removed.

PR #446, opened 2020-09-07
Update bli_type_defs.h

devinamatthews:
Set BLIS_NUM_ARCHS automatically by inserting it as the last member in the enum. Also remove `=0` on first enum member since it is guaranteed by the standard. This make is easier and less error-prone to reorder enum entries.

This change was also included in changes I made to #344 and #345 in remerging them, but I thought it important to make this change earlier rather than later.

### DIFF
--- a/frame/include/bli_type_defs.h
+++ b/frame/include/bli_type_defs.h
@@ -985,7 +985,7 @@ typedef enum
 typedef enum
 {
 	// Intel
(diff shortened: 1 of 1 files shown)


fgvanzee:
This PR has gotten stale and non-trivial to merge, and so I'm going to manually apply and commit its substance outside of this PR.

PR #450, opened 2020-09-29
Add POWER10 support to BLIS

nicholaiTukanov:
This patch adds POWER10 support to BLIS. It also adds two POWER10 microkernels: DGEMM & SGEMM. These kernels are 8x8 and 8x16 respectively.

The kernels utilize the new MMA feature that was introduced in POWER ISA 3.1.


fgvanzee:
This looks good to me. Thanks Nicholai.

PR #471, opened 2021-01-08
Fix kernel-to-config mapping for intel64

### LINKED ISSUE #470: "Illegal Instruction" on intel64
I compiled BLIS using the `intel64` option, but when I run tests with
`make check`, I get the following output:
```
Linking test_libblis.x against 'lib/intel64/libblis.a  -lm -lpthread -lrt'
Running test_libblis.x (fast) with output redirected to 'output.testsuite'
bash: line 2: 50113 Illegal instruction     (core dumped) ./test_libblis.x -g ./testsuite/input.general.fast -o ./testsuite/input.operations.fast > output.testsuite
make: *** [Makefile:834: testsuite-run-fast] Error 132
```

My machine is running
- Bash 5.0
- Make 4.2
- Python 3.8
- Ubuntu 20.04

with a 7th gen Intel i7 processor.

devinamatthews:
Choose last sub-config in the kernel-to-config map if the config list doesn't contain the name of the kernel set. E.g. for "zen: skx knl haswell" pick "haswell" instead of "skx" which was chosen previously. Fixes #470.

### DIFF
--- a/configure
+++ b/configure
@@ -2593,11 +2593,11 @@ main()
 
 			reducedclist="${kernel}"
 
-		# Otherwise, use the first name.
+		# Otherwise, use the last name.
 		else
 
-			first_config=${configs%% *}
-			reducedclist="${first_config}"
+			last_config=${configs##* }
+			reducedclist="${last_config}"
 		fi
 
 		# Create a new "kernel:subconfig" pair and add it to the kconfig_map


devinamatthews:
Tested locally and works.

fgvanzee (commented):
This seems fine for now. Thanks, Devin.

PR #476, opened 2021-01-25
Typed API for GEMMT Inconsistent with Document?

xrq-phys:
Hi.

I suspect that in `frame/3/bli_3_tapi.c` macro generation for typed-API `bli_?gemmt` should be placed in line 410 (after `syr2k`) instead of line 103 (after `gemm`) so that the API generated would be consistent with the document.

BTW the doc says that `syr2k` and `her2k` has `transab` as parameters but actually in the code they're separated as `transa` and `transb`. As typed-API wraps around object-API, I guess the latter should be correct?

### DIFF
--- a/docs/BLISTypedAPI.md
+++ b/docs/BLISTypedAPI.md
@@ -1295,7 +1295,8 @@ where C is an _m x m_ Hermitian matrix stored in the lower or upper triangle as 
 void bli_?her2k
      (
        uplo_t  uploc,
-       trans_t transab,
+       trans_t transa,
+       trans_t transb,
        dim_t   m,
        dim_t   k,
        ctype*  alpha,
@@ -1307,9 +1308,9 @@ void bli_?her2k
 ```
 Perform
 ```
-  C := beta * C + alpha * transab(A) * transab(B)^H + conj(alpha) * transab(B) * transab(A)^H
+  C := beta * C + alpha * transa(A) * transb(B)^H + conj(alpha) * transb(B) * transa(A)^H
 ```
-where C is an _m x m_ Hermitian matrix stored in the lower or upper triangle as specified by `uploc` and `transab(A)` and `transab(B)` are _m x k_ matrices.
+where C is an _m x m_ Hermitian matrix stored in the lower or upper triangle as specified by `uploc` and `transa(A)` and `transb(B)` are _m x k_ matrices.
 
 **Note:** The floating-point type of `beta` is always the real projection of the floating-point types of `A` and `C`.
 
@@ -1371,7 +1372,8 @@ where C is an _m x m_ symmetric matrix stored in the lower or upper triangle as 
 void bli_?syr2k
      (
        uplo_t  uploc,
-       trans_t transab,
+       trans_t transa,
+       trans_t transb,
        dim_t   m,
        dim_t   k,
        ctype*  alpha,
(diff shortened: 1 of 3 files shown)


xrq-phys:
Sorry I made a mistake. `gemmt` and `syr2k` are different by one transpose.

devinamatthews:
No, I think you're right that the gemmt and syr2k interfaces are exactly the same.

devinamatthews:
Ditto docs should have separate transa/transb for her2k/syr2k. Could you add this to the PR?

rvdg:
If I recall correctly, the idea is that traditionally syr2k supports $ C = \alpha A B ^H + B A^H $  and $ C = \alpha A^H B + B^H  $.
The BLIS API also supports $ C = A B + B^H A^H $ and $ C = A^H B^H + B A $.

devinamatthews:
The code indicates that `C = A B^H + A^H B` and `C = A^H B + A B^H` are also supported.

xrq-phys:
On size calculation, `gemmt` assumes `B` to have leading dimension contracted but `syr2k` assumes `B`'s trailing dimension contracted. So maybe it still needs to be separated..

Or we use `INSERT_GENTFUNC_BASIC` instead of `INSERT_GENTFUNC_BASIC` and branch according to `strucc`?

devinamatthews:
Yes, good catch. Simplest is to just keep the same prototype (this is still the same) but copy-paste the definition and modify accordingly.

devinamatthews:
For the record I don't like the "built-in" transpose of B in syr2k, but I guess its about 30 years to late for that objection.

xrq-phys:
`skr2k`'s setting A and B to the same size might provide convenience in some cases but does become confusing when we compare it with `gemm` 🤨. Ummm..

xrq-phys:
Anyone interested in reviewing this PR?
There might be other ones willing to use `gemmt`, I guess.

fgvanzee:
@xrq-phys Thank you for catching the inconsistency between the API documentation for `her2k`/`syr2k` and their actual typed APIs.

> No, I think you're right that the gemmt and syr2k interfaces are exactly the same.

`gemmt` is fundamentally a rank-k update. `her2k` and `syr2k` are fundamentally rank-2k updates. Implementing `gemmt` in terms of `her2k`/`syr2k` will do the wrong thing.

If you remove all the changes in `frame/3/bli_l3_tapi.c` and `.h`, I'll merge the doc fixes.

devinamatthews:
@fgvanzee 'gemmt' and 'syr2k/her2k' DO have the same interface: 'gemmt' simply does 1/2 of the work that `syr2k/her2k' do because we know that the result of A*B is symmetric a priori.

devinamatthews:
The semantics are slightly different, though: `syr2k/her2k` assume `op(A)*op(B^T)` while `gemm` assumes `op(A)*op(B)`.

fgvanzee:
Apologies for the confusion, I misunderstood what the status quo was. (And I really have trouble catching up on long threads like this.)

I see what is being proposed now. I think the proposed changes to `frame/3/bli_l3_tapi.c` and `.h` are appropriate.

fgvanzee:
> `skr2k`'s setting A and B to the same size might provide convenience in some cases but does become confusing when we compare it with `gemm` raised_eyebrow. Ummm..

Instead of focusing on A and B, it may help to start by considering the size of C in either case (*m x n* or *m x m*), and then factor in the *k* dimension, and then scale by 1/2 in the case of `gemmt`: *2mnk* flops for `gemm` vs. *mmk* for `gemmt`.

devinamatthews:
I think what @xrq-phys is saying is that because `syr2k` is `op(A)*op(B^T)`, then both `A` and `B` are `m*k`, whereas in `gemmt` they are `m*k` and `k*m`.

xrq-phys:
That's right.

Anyway. The API standard is consistent in its own sense.

Thanks for merging!

fgvanzee:
@xrq-phys You are welcome! Thanks for your contribution.

PR #482, opened 2021-02-18
Fixed double free in level1v example and freed a

Runkli:
Pointer z was being freed twice and causing a crash.
Pointer a remained after the program terminates.


fgvanzee:
Thank you, @Runkli! Not sure how this slipped by me.

PR #487, opened 2021-03-17
Fix calculation of new pb size

nicholaiTukanov:
This PR will fix a bug in the calculation of the new PB size for I8/I4 GEMM.


fgvanzee:
Is this ready to go, @nicholaiTukanov?

nicholaiTukanov:
@fgvanzee yep

PR #488, opened 2021-03-30
Fix bug in storing POWER10 DGEMM microkernel result

nicholaiTukanov:
There was a bug in the DGEMM P10 kernel that did not store the result of the microkernel correctly. The behavior of the bug was that when the result was being stored, it stored further than the allocated amount for the matrix buffer. This caused the user to see an error when using free(). The reason was due to the wrong index calculation 😞.

### DIFF
--- a/kernels/power10/3/bli_dgemm_power10_mma.c
+++ b/kernels/power10/3/bli_dgemm_power10_mma.c
@@ -168,25 +168,25 @@ void bli_dgemm_power10_mma_8x8
     // handle beta cases
     if (beta_ != 0.0)
     {
-        SAVE_ACC(dv4sf_t, &acc0, rs_c, 0      );
-        SAVE_ACC(dv4sf_t, &acc1, rs_c, 4      );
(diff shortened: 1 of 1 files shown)


nicholaiTukanov:
@fgvanzee This is good to be merged. Apologies for introducing the bug in the first place.

fgvanzee:
@devinamatthews Could you look at the Travis CI failure on this merge. It's obviously unrelated to Nicholai's patch, but I think the SDE stuff might have broken.

devinamatthews:
Looks like we need to "upgrade" to a new version of SDE. I'll see if I can find a new magic filename.

devinamatthews:
Do you want me to make a commit ot fix that here or in another PR?

devinamatthews:
Uhh, I guess I can't here. I'll open a PR.

fgvanzee:
Thanks, @devinamatthews.

@nicholaiTukanov I'll go ahead and merge this given that we're certain the Travis failure has nothing to do with your patch.

PR #489, opened 2021-03-30
Update do_sde.sh

devinamatthews:
Update to a newer version of SDE, and do a direct download as it seems you don't have to click-through the license anymore.


fgvanzee:
Despite the temporary disruption on our end, it looks like Intel moved things in the right direction here.

fgvanzee (approved):

PR #490, opened 2021-04-07
Additional A64fx Comments

xrq-phys:
- Reason for ARMPL's missing data;
- Additional envs / flags for kernel selection;
- Update BLIS SRC commit.


xrq-phys:
In #424 I added a switch to enable/disable registration of `zgemm_nat` kernels.
But maybe it's better to modify `test/3/test_gemm.c` to call `bli_gemm1m`?
Describing the change would be kinda twisted though.

Issue #421, opened 2020-07-12
[ARM64] Clobbering x0-x15 Breaks Down on Clang


xrq-phys:
Dear BLIS developers,

This seems a duplicate of #207 . Sorry for having opened another issue.
The current `arm64` kernels and new `armsve` kernel play well with GCC but breaks down under Clang with:
```
kernels/armv8a/3/bli_gemm_armv8a_opt_4x4.c:73:1: fatal error: inline assembly requires more registers than available.
```

Toolchain:
- LLVM 6.0 + Clang *or*
- ARM Allinea Studio 20.1;

*Possible* Reason:
`Number of input registers + Number of clobbers` becomes larger than 30?

Workaround:
Comment out clobbering of `x0-x15`. These general-purpose registers can be discarded w.r.t. the AArch64 calling convention and results of **C** code lying between function entry and `__asm__ volatile` are saved to memory as input registers. There should be no hazard on doing this.
e.g.
```
kernels/armv8a/3/bli_gemm_armv8a_asm_d6x8.c:1068:
 :// Register clobber list
> // "x0", "x1", "x2","x3","x4",
> // "x5", "x6", "x7", "x8",
> // "x9", "x10","x11","x12",
> // "x13","x14","x15",
  "x16","x17","x18","x19",       
  "x20","x21","x22","x23",
  "x24","x25","x26","x27",
  "v0", "v1", "v2", "v3",
  "v4", "v5", "v6", "v7",
  "v8", "v9", "v10","v11",
  "v12","v13","v14","v15",
  "v16","v17","v18","v19",
  "v20","v21","v22","v23",
  "v24","v25","v26","v27",
  "v28","v29","v30","v31"
...
kernels/armv8a/3/bli_gemm_armv8a_asm_d6x8.c:2061:
:// Register clobber list
> //"x0","x1","x2","x3",
> //"x4","x5","x6",
> //"x7","x8","x9",
> //"x10","x11","x12","x13","x14",
  "x16","x17",
  "x20","x21","x22","x23","x24","x25","x26",
  "x27",       
  "v0","v1","v2",
  "v3","v4","v5",
  "v6","v7","v8",
  "v9","v10","v11",
  "v12","v13","v14",
  "v15","v16","v17","v18","v19",
  "v20","v21","v22","v23",
  "v24","v25","v26","v27",
  "v28","v29","v30","v31"

```

devinamatthews:
[...]

Issue #207, opened 2018-05-14
Unable to build using Android NDK + CLANG


sandwichmaker:
I am interested in building BLIS on android and comparing its performance with OpenBLAS and Eigen.

Even though current versions of the android-ndk ship with the GCC compiler as an option, the use of GCC in the NDK is deprecated:

[https://android.googlesource.com/platform/ndk.git/+/master/CHANGELOG.md](https://android.googlesource.com/platform/ndk.git/+/master/CHANGELOG.md)

Currently the BUILD build system when building for ARM explicitly checks for GCC and fails without it.  

But it is straightforward to modify 

config/arm64/make_defs.mk
config/cortexa57/make_defs.mk

to disable the check and modify the compiler flags to work with CLANG. 

At this point we hit a more interesting hurdle

kernels/armv8a/3/bli_gemm_armv8a_opt_4x4.c

which fails with the CLANG complaining 

kernels/armv8a/3/bli_gemm_armv8a_opt_4x4.c:73:1: fatal error: inline assembly requires more registers than available

Poking at the clobber list, it seems the function is using 9 general purpose registers too many. Commenting out any nine general purpose registers makes the function compile (incorrectly of course) and then fail on the next function in the file. 

This is true for NDK version r15c, r16b and r17.

Is this something that can be easily fixed? Also I am curious as to why BLIS is using these rather large functions as inline assembly and not just assembler files?

I am happy to provide more detailed reproduction instructions if that will help.

sandwichmaker:
Passing -fno-integrated-as as recommended by [Clang Migration Notes](https://android.googlesource.com/platform/ndk.git/+/master/docs/ClangMigration.md) does not help either.

fgvanzee:
@sandwichmaker Thanks for this report.

Unfortunately we are light on ARM expertise in the core BLIS development group, so I can't speak for why the code was written to clobber the nine extra registers.

We just added a feature that allows us to modify the default compiler search list based on OS (specifically, the result of `uname -s`) and also fixed a related issue with the compiler being miscommunicated to `make` from `configure`. If you can tell me what `uname -s` returns on ARM, we might be able to isolate your environment and change the search order accordingly, and then you will not have to mess with setting the compiler in `make_defs.mk`.

sandwichmaker:
@fgvanzee this is not a compiler search issue. I am cross compiling for android on linux using the [android-ndk](https://developer.android.com/ndk/downloads/). It is using the right compiler toolchain. 

[...]

PR #492, opened 2021-04-14
Allow clang for ThunderX2 config

devinamatthews:
Needed for compiling on e.g. Mac M1. AFAIK clang supports the same -mcpu flag for ThunderX2 as gcc.

### DIFF
--- a/config/thunderx2/make_defs.mk
+++ b/config/thunderx2/make_defs.mk
@@ -65,7 +65,11 @@ CKOPTFLAGS     := $(COPTFLAGS) -O3 -ftree-vectorize
 ifeq ($(CC_VENDOR),gcc)
 CKVECFLAGS     := -mcpu=thunderx2t99
 else
-$(error gcc is required for this configuration.)
+ifeq ($(CC_VENDOR),clang)
+CKVECFLAGS     := -mcpu=thunderx2t99
+else
+$(error gcc or clang is required for this configuration.)
+endif
 endif
 
 # Flags specific to reference kernels.


figual:
Issues #207 and #421 will rise compiling ARM64 microkernels with clang. I am working on reducing register pressure to solve the issue precisely to compile for M1 if you are interested. You can install gcc in the meanwhile :)

devinamatthews:
I'm told homebrew isn't production-ready on M1. What is the recommended way to install gcc?

jdiamondGitHub:
> Issues #207 and #421 will rise compiling ARM64 microkernels with clang. I am working on reducing register pressure to solve the issue precisely to compile for M1 if you are interested. You can install gcc in the meanwhile :)

In reviewing the clang gnu assembly issue, are you saying there is no way for clang assembly to access half the registers?  Or does clang just have a different syntax that we could use and thereby convert the kernel to a format that would work?  (Porting kernels to M1) Thanks.

jdiamondGitHub:
test post

jdiamondGitHub:
Figual, is your experience that there is no way clang inline assembly can access half the registers on ARM?  Or could we port to clang just by rewriting the assembly in a different syntax?  Thanks.

devinamatthews:
@jdiamondGitHub I don't think it's that clang can only access half the registers, just that for some reason the compiler keeps one of the registers for itself (like %rbp is sometimes reserved in x86).

jdiamondGitHub:
Thanks, Devin.  That's the other point I wanted to raise:  We know that 
the x86 kernels had to be modified for Mac due to the difference in ABI 
(e.g., which register was used for the stack, etc.)  Does anyone know 
how Apple's use of ARM registers differs from Linux?  I'm thinking 
that's another reason we might have to modify the ARM kernels for Macs.

jdiamondGitHub:
The main point here being that even if we set up gcc on Mac, the kernel 
might not work.  It might crash the machine, similar to what happened 
before we modified the x86 kernels for Mac.

jdiamondGitHub:
Looks like maybe X18 is the issue?
https://developer.apple.com/documentation/xcode/writing_arm64_code_for_apple_platforms

jdiamondGitHub:
Folks, I don't think the gcc approach will work.  I was just Googling 
it, and gcc has not yet created an M1 back end, implying that anything 
gcc produces will also not respect the Apple specific registers.  So the 
issue with using gcc is not just about the kernel, but with blis itself.

Seems like we have to focus on making the ARM kernels compatible with 
Clang and the M1 reserved registers.

xrq-phys:
Folks,

I'd like to suggest that rather than using ThunderX config for Arm Mac, let's init a new config for Apple since cache size of Firestorm & Icestorm is much larger than those of Cortex-A57.

This is a simple comparison between config `cortexa57` (roughly the same as `thunderx2`) and a slightly tuned one:
output_st_dgemm_asm_blis_firestorm_tag0

PR #498, opened 2021-05-05
Replace bli_dlamch with something less archaic

### LINKED ISSUE #497: Issues with  ?lamch routines: use O0 to compile them and static variables
It seems that BLIS compiles slamch and dlamch using optimization flags enabled.
My configuration is

`./configure --enable-blas --enable-cblas --enable-pba-pools --enable-sba-pools  --enable-threading=openmp --prefix=$HOME/amd/src/local zen2`

and the compilation of these routines is

```
[zampins@cn514-15-r blis]$ grep lamch.c make2.log
/sw/csi/gcc/10.2.0/bin/gcc -O3 -Wall -Wno-unused-function -Wfatal-errors -fPIC -std=c99 -D_POSIX_C_SOURCE=200112L -fopenmp -Iinclude/zen2 -I./frame/3/ -I./frame/ind/ukernels/ -I./frame/3/ -I./frame/1m/ -I./frame/1f/ -I./frame/1/ -I./frame/include -DBLIS_VERSION_STRING=\"3.0.0\" -DBLIS_IS_BUILDING_LIBRARY -fvisibility=hidden -c frame/base/noopt/bli_dlamch.c -o obj/zen2/frame/base/noopt/bli_dlamch.o
/sw/csi/gcc/10.2.0/bin/gcc -O3 -Wall -Wno-unused-function -Wfatal-errors -fPIC -std=c99 -D_POSIX_C_SOURCE=200112L -fopenmp -Iinclude/zen2 -I./frame/3/ -I./frame/ind/ukernels/ -I./frame/3/ -I./frame/1m/ -I./frame/1f/ -I./frame/1/ -I./frame/include -DBLIS_VERSION_STRING=\"3.0.0\" -DBLIS_IS_BUILDING_LIBRARY -fvisibility=hidden -c frame/base/noopt/bli_slamch.c -o obj/zen2/frame/base/noopt/bli_slamch.o

```

Using gcc 10.2 with `-march=znver2` on an AMD EPYC 7702 let my code (compiled with `-fopenmp`) emit this warning

```
(// WARNING. The value EMIN may be incorrect:-,  EMIN = ,i8,/ If, after inspection, the value EMIN looks, acceptable please comment out ,/ the IF block as marked within the code of routine, DLAMC2,,/ otherwise supply EMIN explicitly.,/)(// WARNING. The value EMIN may be incorrect:-,  EMIN = ,i8,/ If, after inspection, the value EMIN looks, acceptable please comment out ,/ the IF block as marked within the code of routine, DLAMC2,,/ otherwise supply EMIN explicitly.,/)
```

I'm not familiar with BLIS makefile rule defintions to be able to provide a fix for this. However, I see the ?lamch routines are in `frame/base/noopt/` so there's probably hope that these routines will be compiled with `-O0`. Same happens with the amd fork of blis as well as with the libflame code. Should I submit an issue for libflame too or this one is enough?

devinamatthews:
Since it's not 1970 anymore the standard library helpfully provides all of the floating point limit constants needed for `[sd]lamch`. Fixes #497.


fgvanzee:
I seem to recall trying this many years ago, and, IIRC, I found that the more modern/straightforward implementation of dlamch caused issues somewhere down the line, probably in the LAPACK test suite.

But before we can even consider its effects on LAPACK, we should get the patch working with our own test suite. :)

devinamatthews:
Yeah, that should teach me not to author commits in the browser. I'll get it to compile and pass checks locally. Re LAPACK tests, is there an easy way to check compliance (e.g. putting this in libflame and running the LAPACK testsuite)? I can certainly print out the old and new values on a few archs and see where the differences lie. In particular, I think the "safe minimum" values may be slightly less than `[FLT|DBL]_MIN` when denormals are used.

devinamatthews:
BLAS tests pass now.

fgvanzee:
> Yeah, that should teach me not to author commits in the browser. I'll get it to compile and pass checks locally.

It happens. At least it was just a branch.

> Re LAPACK tests, is there an easy way to check compliance (e.g. putting this in libflame and running the LAPACK testsuite)?

I think the easiest way is to download netlib LAPACK and plug it in. (There should be a place to specify the BLAS library for testing purposes.) libflame's LAPACK code hasn't been refreshed in some time, so I don't think that's the best vehicle for testing anymore.

> I can certainly print out the old and new values on a few archs and see where the differences lie. In particular, I think the "safe minimum" values may be slightly less than `[FLT|DBL]_MIN` when denormals are used.

Thanks, that would be very interesting to see.

No matter what your initial tests show, though, I'm nervous enough about this in the short term that I think you should add your new implementation as a preprocessor branch (guarded by something like `BLIS_ENABLE_STD_LAMCH`) so that it can be switched at compile time. (No need to create a configure switch yet; we'll consider this an expert option for now, similar to `BLIS_DISABLE_BLAS_DEFS`.) This will allow us to gain confidence in the newer implementation while preserving the option of quickly falling back to the older code.

devinamatthews:
@fgvanzee for IEE754 floating point all values match the "legacy" `_lamch`. What systems that BLIS supports could possibly do anything funny?

devinamatthews:
Looking at the LAPACK version, I'm actually convinced that, with the `sfmin` tweak I added, the "new" and "old" versions will always agree.

fgvanzee:
> @fgvanzee for IEE754 floating point all values match the "legacy" `_lamch`. What systems that BLIS supports could possibly do anything funny?

No idea. I have a distant memory of looking at this maybe 5-6 years ago, so perhaps it was all due to a bug that got fixed. (I mean, I don't *want* them to be different. So I'm glad you're seeing consistent numbers now.)

Nevertheless, I don't see any harm in keeping the old code disabled via a preprocessor guard.

devinamatthews:
Preprocessor guard added. Define `BLIS_ENABLE_LEGACY_SLAMCH` to use the **old** code.

fgvanzee:
Thanks, much better.

Out of curiosity, why does the new `bli_dlamch()` sometimes use the `FLT_` constants?
```c
		case 'E': return DBL_EPSILON;
		case 'S': return safe_min;
		case 'B': return FLT_RADIX;
		case 'P': return FLT_RADIX*DBL_EPSILON;
		case 'N': return DBL_MANT_DIG;
		case 'R': return FLT_ROUNDS == FE_TONEAREST ? 1.0 : 0.0;
		case 'M': return DBL_MIN_EXP;
		case 'U': return DBL_MIN;
		case 'L': return DBL_MAX_EXP;
		case 'O': return DBL_MAX;
```

fgvanzee:
> Out of curiosity, why does the new `bli_dlamch()` sometimes use the `FLT_` constants?

@devinamatthews In case you missed my question from earlier.

devinamatthews:
I could have sworn I responded... oh well. This is because things like the radix and rounding behavior don't depend on the precision, so there is only one constant instead of three (float, double, long double).

fgvanzee:
Okay, thanks.

Issue #453, opened 2020-11-02
Migrate Travis CI testing from Trusty to Xenial or Bionic


fgvanzee:
Update `.travis.yml` file to use Ubuntu 16.04 (Xenial) or 18.04 (Bionic). The config file currently specifies use of 14.04 (Trusty). See [1] for migration guide and [2] for a list of supported Ubuntu distributions.

[1] https://docs.travis-ci.com/user/trusty-to-xenial-migration-guide
[2] https://docs.travis-ci.com/user/reference/linux

devinamatthews:
Addressed by #500.

PR #500, opened 2021-05-19
Upgrade Travis CI for Arm SVE

xrq-phys:
cf. comments in #424 , this PR is to:

- Include `armsve` kernels and the `armsve` configuration to Travis CI tests.

For this reason, the following changes are done in the Travis CI config:

- Upgraded linux ver. from `trusty` to `focal` because `armsve` kernels requires GCC 10+ to compile and QEmu 4.4+ to emulate.
- Changed several toolchains used by Travis CI because previous toolchain is no longer available on `focal`.

To-dos:

- ~Travis CI seems also supporting Arm nodes.
   Compared to x86_64+QEmu, maybe using Arm64+ArmIE a better way for testing Arm kernels?
   At least `armv8a` and `armv7a` can be tested without emulation.
   ArmIE's better stability over QEmu can be another advantage.~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,14 @@ matrix:
   # clang build
   - os: linux
     compiler: clang
-    env: OOT=0 TEST=0 SDE=0 THR="none" CONF="auto" \
-      PACKAGES="clang-8 binutils"
+    env: OOT=0 TEST=0 SDE=0 THR="none" CONF="auto"
+      # There seems to be some difficulty installing 2 Clang toolchains of different versions.
+      # Use the TravisCI default.
+      # PACKAGES="clang-8 binutils"
   # macOS with system compiler (clang)
   - os: osx
     compiler: clang
-    env: OOT=0 TEST=1 SDE=0 THR="none" CONF="auto" \
-      PACKAGES="clang-8 binutils"
+    env: OOT=0 TEST=1 SDE=0 THR="none" CONF="auto"
   # cortexa15 build and fast testsuite (qemu)
   - os: linux
     compiler: arm-linux-gnueabihf-gcc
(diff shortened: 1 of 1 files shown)


xrq-phys:
BTW, I'm not sure about benefits but by switching from travis-ci.org to travis-ci.com we will be able to use `arm64-graviton2` in addition to `arm64`.

devinamatthews:
@xrq-phys how did you make the PR a draft? I've never been able to find that button or whatever it is.

.travis.yml
```diff
@@ -43,38 +53,33 @@ matrix:
...
-    - gcc-6
-    - binutils-2.26
-    - clang
+- if [ "$CC" = "gcc"  ] && [ "$TRAVIS_OS_NAME" = "linux" ]; then export CC="gcc-8"; fi
+- if [ -n "$PACKAGES" ] && [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install -y $PACKAGES; fi
 script:
 - export DIST_PATH=.
-- pwd
 - if [ $OOT -eq 1 ]; then export DIST_PATH=`pwd`; mkdir ../oot; cd ../oot; chmod -R a-w $DIST_PATH; fi
+- $DIST_PATH/configure -t $THR CC=$CC CFLAGS=$CFLAGS $CONF
```
devinamatthews, line 76: Why add `CFLAGS=$CFLAGS`? This doesn't seem to get set to anything special.

devinamatthews (approved):
Looks good, just one comment below.

devinamatthews:
I would leave the arm64 HW support for another PR. We would probably want to wait until after migrating to travis-ci.com for that anyways so graviton2 can be included.

.travis.yml
xrq-phys: This is from the old `.travis.yml`.
BLIS' `configure` is not Autoconf so I thought influence of args could be somehow different from env.

If it is not the case, I guess it's enough to just: `$DIST_PATH/configure -t $THR $CONF` ?

devinamatthews: The old line 74 doesn't have CFLAGS (or at least, what I see here in the diff on the LHS doesn't). I would remove it for consistency.

xrq-phys: Sorry.
The old `.travis.yml` has only `CC=$CC`.

I used `CFLAGS=-DSVE_XXX` upon other configs but now CFLAGS is not set anyway so this is an unnecessary extra by myself. Will remove.

xrq-phys:
Got it.

I've set up travis.com build on my fork so hopefully native arm64 support (maybe ppc also?) will be there as soon as master's CI moves :)

fgvanzee:
@xrq-phys We have credits ready to go for travis-ci.com, but we are delaying the migration for now because it requires us to click a button that say we are using the "beta" version of the service. (I asked their support about this, and they basically said, "Oh, don't worry about that. It's full service." I did not find their reply super convincing, although it's better than silence I suppose.)

devinamatthews:
@fgvanzee this PR looks good to me. I'll merge if you're OK with it (note the travis.com-related stuff @xrq-phys mentioned is **not** in this PR).

fgvanzee:
> (note the travis.com-related stuff @xrq-phys mentioned is not in this PR).

Understood.

It looks fine to me.

fgvanzee:
Also, it appears that this PR obviates #453.

PR #506, opened 2021-05-27
BLIS on Darwin_Aarch64

xrq-phys:
Solved: Assembler of Apple on Arm64 seems to require additional "tagging" for symbols to be identified as "local".

~Still, compilation requires GCC because of #421.~
Solved: Clang requires kernel assemblies to clobber less registers.

~Keeping this a draft as we might have some new Arm64+NEON kernels in a close future.~

### DIFF
--- a/kernels/armv8a/3/armv8a_asm_utils.h
+++ b/kernels/armv8a/3/armv8a_asm_utils.h
@@ -1,0 +1,49 @@
+/*
+
+   BLIS
+   An object-based framework for developing high-performance BLAS-like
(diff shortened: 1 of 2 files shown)


xrq-phys:
Hi.

I checked the 2014 ArmV8A kernels and noticed many registers able for reusing.
Renaming scheme is described in commit message of 916e1fa and the d6x8 kernels can now pass Clang compilation.

This should directly address #492.

devinamatthews:
Awesome!

PR #515, opened 2021-07-02
Fixed configure script bug.

chengguosun:
Details:
- Fixed kernel list string substitution error by adding function substitute_words in configure script.
- For example: zen in zen2 would be replaced by other strings, resulting in the wrong name and not containing the correct kernel header file.

### DIFF
--- a/configure
+++ b/configure
@@ -775,7 +775,8 @@ read_registry_file()
 					# canonicalize whitespace, and then remove duplicate kernel
 					# set names, if they exist. Finally, update the kernel registry
 					# with the new kernel list.
-					newklist=$(echo -e "${klisttmp}" | sed -e "s/${ker}/${kers_ker}/g")
+					#newklist=$(echo -e "${klisttmp}" | sed -e "s/${ker}/${kers_ker}/g")
+                    newklist=$(substitute_words "${ker}" "${kers_ker}" "${klisttmp}")
 					newklist=$(canonicalize_ws "${newklist}")
 					newklist=$(rm_duplicate_words "${newklist}")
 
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -797,6 +798,25 @@ read_registry_file()
 	done
 }
 
+substitute_words()
+{
+	local word new_words list
+	
+	word=$1
+	new_words=$2
+	list=$3
```
devinamatthews, line 807: Should these three lines have double quotes?

configure
```diff
@@ -797,6 +798,25 @@ read_registry_file()
 	done
 }
 
+substitute_words()
+{
+	local word new_words list
```
devinamatthews, line 803: Shouldn't `newlist` also be a local?

configure
chengguosun: You're right, it's better to use double quotes. thanks.

configure
chengguosun: newlist should be local from programming specification, it is careless of me.

devinamatthews:
@fgvanzee reminder update CONTRIBUTIONS if @chengguosun isn't in there already.

fgvanzee:
I guess you mean `CREDITS`? I will.

fgvanzee:
@chengguosun Do you care to have your full name (and/or organization) in the `CREDITS` file? Please let me know your name if you're interested in that. Otherwise I'll just use your handle.

chengguosun:
> @chengguosun Do you care to have your full name (and/or organization) in the `CREDITS` file? Please let me know your name if you're interested in that. Otherwise I'll just use your handle.

@fgvanzee My first name is Chengguo, last name is Sun. I have updated my profile, you can use it. Thanks a lot.

PR #519, opened 2021-07-06
Fix installation from out-of-tree builds

awild82:
# Issue
Currently, installing from out of source builds fails, since the Makefile can't find the `blis.pc.in` file. This can be reproduced with:
```
git clone https://github.com/flame/blis.git
mkdir blis-build && cd blis-build
../blis/configure x86_64
make
make install
```
which will fail with a message such as:
```
make: *** No rule to make target `blis.pc.in', needed by `<path/to/install/location>/share/pkgconfig'.  Stop.
```

# Fix
This PR fixes this issue by creating a symlink to `blis.pc.in` in the case of an out of source build in the same fashion as the symlinks for the `Makefile` and `common.mk` files.

### DIFF
--- a/configure
+++ b/configure
@@ -3629,6 +3629,23 @@ main()
 			exit 1
 		fi
 
+		# If 'blis.pc.in' symlink does not already exist in the current
+		# directory, create a symbolic link to it. If one does exist, we
+		# use -f to force creation of a new link.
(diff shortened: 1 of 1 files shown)


devinamatthews:
@fgvanzee reminder update CREDITS

PR #524, opened 2021-07-08
Add vzeroupper to Haswell gemm kernels.

### LINKED ISSUE #523: Haswell kernels lost vzeroupper at some point
"Old" haswell kernels has vzeroupper to avoid penalty when mixed with legacy SSE code. For some reason, the current kernels don't have this anymore. AFAICT the compiler does **not** insert this automatically so we should put it back in.

devinamatthews:
Fixes #523.

### DIFF
--- a/kernels/haswell/3/bli_gemm_haswell_asm_d6x8.c
+++ b/kernels/haswell/3/bli_gemm_haswell_asm_d6x8.c
@@ -866,27 +866,27 @@ void bli_sgemm_haswell_asm_6x16
 
 
 
-
 	label(.SDONE)
 
 
-    vzeroupper()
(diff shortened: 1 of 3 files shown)


devinamatthews:
@fgvanzee if you know of a specific reason that vzeroupper was removed let me know, otherwise I'll merge.

devinamatthews:
Looks like the reason is that the zen kernels got copy-pasted from haswell back before this was added, then were co-opted to replace the existing haswell kernels. I also fixed an issue with insufficient prefetch of A that I've fixed previously at least once.

fgvanzee:
I'm making a few whitespace adjustments. Then I'll merge.

PR #528, opened 2021-08-12
Experimental version of obj_t using bit fields for the various flags.

devinamatthews:
- Removed the info and info2 fields and replaced with bitfield entries for each distinct flag/property.
- Simplified and rearranged the bitwise definitions of the various enumerations.
- Removed many unused/underused functions from the obj_t API.

### DIFF
--- a/frame/include/bli_type_defs.h
+++ b/frame/include/bli_type_defs.h
@@ -42,25 +42,6 @@
 // -- BLIS basic types ---------------------------------------------------------
 //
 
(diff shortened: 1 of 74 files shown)


frame/include/bli_type_defs.h
```diff
@@ -571,6 +389,11 @@ typedef enum
```
fgvanzee, line 389: Peas review me for potential non-contiguous enum bug.

PR #531, opened 2021-08-15
L3 oapi without macros

devinamatthews:
Simplify the l3 oapi by removing the rather unnecessary macro layer. This is an example of how many parts of BLIS which are currently macros, but don't really need to be could be potentially improved (I guess it depends on your viewpoint though :smile:).


devinamatthews:
@fgvanzee I think this is actually ready to merge now. All of the L3 operations have the full interface (_ex, nat, ind, etc.) and all of the tests past.

fgvanzee:
I found a regression. Previously, calling `herk` (for example) with a C that is tagged `BLIS_GENERAL` (instead of `BLIS_HERMITIAN`) would have triggered a runtime error. It seems that this code no longer checks for the expected `struc_t` value. Was this intentional?

devinamatthews:
If I reinstate the `_check` routines, is calling them from the oapi layer sufficient?

devinamatthews:
Shouldn't `gemmt` be doing the check on `C`?

fgvanzee:
I'm also seeing that you made a pretty big reorganization of the files that do the expert vs. non-expert API.

I think I'm going to need quite a bit more time to review this.

fgvanzee:
> If I reinstate the `_check` routines, is calling them from the oapi layer sufficient?

I think that would be okay, yes. Was there some other alternative that you had in mind?

devinamatthews:
> > If I reinstate the _check routines, is calling them from the oapi layer sufficient?

> I think that would be okay, yes. Was there some other alternative that you had in mind?

No, I just didn't know if this had to happen in `_front` specifically. I will add the check functions back in. BTW none of this is actually tested in the testsuite :).

fgvanzee:
> > > If I reinstate the _check routines, is calling them from the oapi layer sufficient?
> 
> > I think that would be okay, yes. Was there some other alternative that you had in mind?
> 
> No, I just didn't know if this had to happen in `_front` specifically. I will add the check functions back in. BTW none of this is actually tested in the testsuite :).

I have confirmed that the only thing that happens between `bli_l3_oapi.c` and the calling of the `_front()` functions is getting the context and runtime `structs` squared away.

devinamatthews:
@fgvanzee I added back the error checking.

devinamatthews:
@fgvanzee should be ready to merge.

fgvanzee:
Thanks, I'll start reviewing it.

fgvanzee:
@devinamatthews Turns out we *did* need the `if ( !cntx )` conditional after all, because operations like `herk` call their `_check()` function before the `cntx_t*` is valid (which doesn't happen until `gemmt` is called).

BUT, I decided to solve it by moving the `bli_check_sufficient_stack_buf_size()` call into the gks, since that was where we agreed we would go eventually anyway.

fgvanzee:
I'll put together a squash-merge commit log and then squash-merge later today.

PR #533, opened 2021-08-20
ARMv8 PACKM and GEMMSUP Kernels + Apple Firestorm Subconfig

xrq-phys:
Hi!

This PR adds a subconfig, some supplementary `gemm` sizes, some `packm` kernels and a set of `dgemmsup` kernels.

The subconfig is roughly tuned against Apple's latest CPUs (i.e. CPU part of A14 and M1). This should slightly improve performance on Arm-based mac machines (c.f. #492, *the peak a bit higher than OpenBLAS*).

Regarding `gemmsup`, I did not write assembly for all sizes but relied on some fallback methods:
- `gemmsup_ref` for `rv` cases;
- NEON-intrinsic-based `bli_dgemmsup_rd_armv8a_int_???` for `rd` cases.

image

image

image


devinamatthews:
What shape GEMM is the last graph?

devinamatthews:
@xrq-phys have all of the build problems on M1 been worked out?

xrq-phys:
@devinamatthews Sorry I forgot to add labels there. The last graph is DGEMM M=8 N=6 K=P.

Regarding build problems, LLVM requires number of `__asm__` registers to be limited anyway. Though I cannot guarantee for all cases (no idea how LLVM routes its registers), at least all the assemblies are adjusted to pass Xcode 12.5.

Another option could be GCC. There seemed to be sayings that GCC 10 is not fully trustable on M1, but what about GCC 11?

devinamatthews:
@xrq-phys if it builds, compiles, and runs on M1 without modification then that is awesome. We need to dust off #344 and then everything will look very nice.

devinamatthews:
@fgvanzee can you give this a final blessing?

devinamatthews:
@xrq-phys is this ready to merge?

devinamatthews:
@fgvanzee can you also give this one the OK if @xrq-phys says it's ready?

fgvanzee:
FWIW, I think the conflict in `bli_arch.c` is a result of me merging PR #344 earlier today.

I could try to resolve it by adding a commit to the PR branch, but I will likely have to first figure out these access token things. :confused:

devinamatthews:
I can fix it.

devinamatthews:
Easiest merge ever.

fgvanzee:
Did you do the access token thing?

devinamatthews:
No, I just clicked on "Resolve Conflict". There is a web conflict editor.

fgvanzee:
Ah ha! You cheated! :upside_down_face: 

But it got the job done. Thanks.

devinamatthews:
Converted to draft while working on beta == 0. @xrq-phys please convert back to a regular PR when it's ready for merging.

xrq-phys:
This branch now fully supports `*beta == 0` handling. Marking ready for review :D

devinamatthews:
@fgvanzee the Appeveyor failure is a temporary network problem and not a real failure.

devinamatthews:
There are two little things that I want to add before merging:

1. A Travis test for M1 because there is a lot of new packing and gemmsup code.
2. Autodetection. There is already a "stub" for M1, it just needs to activate the right config.

xrq-phys:
@devinamatthews Oh thanks for the reminder I forgot to push that piece of change.
But in fact there seems a problem with `<sys/sysctl.h>` on Xcode 13.0 as:
```
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/sysctl.h:83:
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/ucred.h:101:2: fatal error: unknown type name 'u_int'
        u_int   cr_version;             /* structure layout version */
```

Would you mind me commenting out that `#include`? It's not used at the moment anyway.

devinamatthews:
Oh yeah, yes it is. I think I put that in there so we could use hwctl but I ended up just hard-coding M1 for now.

xrq-phys:
@devinamatthews For Travis part would a duplicate of `cortexa57` work?

Anyway it's better let you decide. My free Travis-CI has expired already ;)

devinamatthews:
Putting in Travis test now.

xrq-phys:
One CRR case fallen.
Sorry I didn't do BLAS tests.

xrq-phys:
Problematic routines are `bli_dgemmsup_rv_armv8a_asm_8x4m` and `bli_dgemmsup_rv_armv8a_int_6x4mn`.

Trying to fix.

xrq-phys:
@devinamatthews The two kernels are fixed.

Registered subconfig `firestorm` to `arm64` and confirmed (hard-coded) autodetect is working :D.

devinamatthews:
Hmmm. #3 seems to be optional since I am also getting wrong `zgemm1m` for larger sizes too. `make check` doesn't test 1m (which it should!) so Travis wouldn't catch this.

xrq-phys:
Fixed.

This kind of thing worked for `1m`:
```
	bli_cntx_set_l3_nat_ukrs
	(
	  1,
	  // BLIS_GEMM_UKR, BLIS_FLOAT,    bli_sgemm_armv8a_asm_8x12, FALSE,
	  BLIS_GEMM_UKR, BLIS_DOUBLE,   bli_dgemm_armv8a_asm_6x8r, TRUE,
	  cntx
	);
```

It seems that mixing row and column-preferring bulk kernels would break `1m`.
Anyway, moved all row-preferring bulk kernels to `old` since they didn't show much advantage.

devinamatthews:
Confirmed fixed.

Issue #539, opened 2021-09-09
Inline ASM clobber list means no spare register for LLVM


jlinford:
LLVM and Arm Compiler for Linux (ACfL) fail to compile for armsve or a64fx. This code is also failing on upstream LLVM: https://godbolt.org/z/PPzhebnan
```
error: inline assembly requires more registers than available
```
Commenting out Line 27 makes LLVM/ACfL work.

In short, the problem seems to be that there is a long clobber list in the inline assembly, and that there are many function arguments. When you add extra function arguments into the list in the ASM then it means you can include fewer registers in the clobber list, and vice versa.

Affected files:
- https://github.com/flame/blis/blob/master/kernels/armsve/3/bli_gemm_armsve256_asm_d8x8.c
- https://github.com/flame/blis/blob/master/kernels/armsve/3/sup/bli_gemmsup_cv_armsve_asm_d2vx10_unindexed.c
- https://github.com/flame/blis/blob/master/kernels/armsve/3/sup/bli_gemmsup_rv_armsve_asm_d2vx10_unindexed.c

Arm's LLVM and GNU compiler teams have looked into it and concluded that these files are pushing the limits of reasonable use of inline asm.  GCC is only able to compile this because it is making an unsafe assumption that saves it the one additional register (over LLVM) that it needs to be able to compile the input.  GCC is using sp+imm addressing mode, so is using less registers allowing compiler to allocate registers for other expressions.  But inline asm does not guarantee that sp+imm addressing mode is always used in such case, so this code is not portable to other compilers.  It's arguable that GCC works through behavior that could be viewed as a bug.

Recommended solutions:
 1. Converted the affected files to straight assembler.  This seems to be the easiest path forward.
 2. If the files must be implemented in C, re-implement in ACLE intrinsics.

@xrq-phys, @stepannassyr, thoughts?

stepannassyr:
Well great - I've actually been working on something that potentially uses inline asm even more heavily than any kernels I've  manually written so far :D

With Solution 2 tight control over which architectural register gets used is lost and it's also incompatible with my current work.

I think Solution 1 is the way to go, but it does mean some rework is required. I'm actually mostly using inline asm to avoid the asm boilerplate around stuff like functions, but I'm also not a big fan of relying on bugs/undefined behaviour. While in the context of the upstreamed code only a couple files need to be changed to asm, for my development code and future work quite a bit more needs reworking, so yeah... I agree this is the best solution but I currently cannot invest the time.

[...]

PR #540, opened 2021-09-12
Arm SVE Fix Clang Compilation & Cleanup Experimental Clutter

xrq-phys:
This should address #539 .

Detailed strategy is described in thread discussion of the issue.


kernels/armsve/3/bli_gemm_armsve256_asm_d8x8.c
```diff
@@ -60,6 +60,10 @@ void bli_dgemm_armsve256_asm_8x8
 {
 	void* a_next = bli_auxinfo_next_a( data );
 	void* b_next = bli_auxinfo_next_b( data );
+	static int called = 0;
+	if (!called)
+		fprintf(stderr, "8x8 called\n");
+	called = 1;
```
devinamatthews, line 66: This looks like some leftover debugging stuff?

kernels/armsve/3/bli_gemm_armsve256_asm_d8x8.c
```diff
@@ -70,35 +74,35 @@ void bli_dgemm_armsve256_asm_8x8
...
+// " ldr x7,%[alpha]                            \n\t" // Alpha address      
+// " ldr x8,%[beta]                             \n\t" // Beta address      
+// "                                            \n\t" 
+// " ldr x9,%[cs_c]                             \n\t" // Load cs_c
+// " ldr x13,%[rs_c]                            \n\t" // Load rs_c.
+"                                            \n\t"
+" lsl x10,%9,#3                              \n\t" // cs_c * sizeof(double)
+" lsl x14,%10,#3                             \n\t" // rs_c * sizeof(double). 
+"                                            \n\t"
+" add x20,%2,x10                             \n\t" //Load address Column 1 of C
```
devinamatthews, line 97: I think I'm going to push a patch to use asm names (e.g. %[c]) for whoever ends up maintaining this in the future.

devinamatthews (commented):
I made a couple of comments in the code. Is `bli_gemm_armsve256_asm_d8x8` actually used?

devinamatthews:
@xrq-phys if that kernel isn't currently used then I'll put it in the "old" folder instead of fixing it up.

devinamatthews:
@xrq-phys I checked and indeed they are not used. I prefer to just move all the old stuff out of the way, which I have done in #554. Thanks for the fix though, this is definitely the way to go for future kernels.

PR #545, opened 2021-09-20
bli_error: more cleanup on the error strings array

hominhquan:
- There was redundance between the macro BLIS_MAX_NUM_ERR_MSGS (=200) and
  the enum BLIS_ERROR_CODE_MAX (-170), while they both mean the same thing:
  the maximal number of error codes/messages.
- The previous initialization of error messages at compile time ignored that
  the 'bli_error_string' array still occupies useless memory due to 2D char[][]
  declaration. Instead, it should be just an array of pointers, pointing at
  strings in .rodata section.
- This commit does the two modifications:
   * retired macros BLIS_MAX_NUM_ERR_MSGS and BLIS_MAX_ERR_MSG_LENGTH everywhere
   * switch bli_error_string from char[][] to char *[] to reduce its footprint
     from 40KB (200*200) to 1.3KB (170*sizeof(char*)).
     (No problem to use the enum BLIS_ERROR_CODE_MAX at compile-time,
     since compiler is smart enough to determine its value is 170.)

### DIFF
--- a/frame/base/bli_error.c
+++ b/frame/base/bli_error.c
@@ -36,7 +36,7 @@
 #include "blis.h"
 
 // Internal array to hold error strings.
-static char bli_error_string[BLIS_MAX_NUM_ERR_MSGS][BLIS_MAX_ERR_MSG_LENGTH] =
+static char *bli_error_string[-BLIS_ERROR_CODE_MAX] =
 {
 	[-BLIS_INVALID_ERROR_CHECKING_LEVEL]         = "Invalid error checking level.",
 	[-BLIS_UNDEFINED_ERROR_CODE]                 = "Undefined error code.",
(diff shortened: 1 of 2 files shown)


nibrunie:
> * There was redundance between the macro BLIS_MAX_NUM_ERR_MSGS (=200) and
>   the enum BLIS_ERROR_CODE_MAX (-170), while they both mean the same thing:
>   the maximal number of error codes/messages.
> * The previous initialization of error messages at compile time ignored that
>   the 'bli_error_string' array still occupies useless memory due to 2D char[][]
>   declaration. Instead, it should be just an array of pointers, pointing at
>   strings in .rodata section.
> * This commit does the two modifications:
>   
>   * retired macros BLIS_MAX_NUM_ERR_MSGS and BLIS_MAX_ERR_MSG_LENGTH everywhere
>   * switch bli_error_string from char[][] to char _[] to reduce its footprint
>     from 40KB (200_200) to 1.3KB (170_sizeof(char_)).
>     (No problem to use the enum BLIS_ERROR_CODE_MAX at compile-time,
>     since compiler is smart enough to determine its value is 170.)

It looks like you are aggressively trying to save code size.

hominhquan:
> It looks like you are aggressively trying to save code size.

Yes, and it will not stop here ;)

Issue #532, opened 2021-08-19
Having trouble in building with cross-compiler


jongwonjlee:
Hi, I appreciate your hard work! I'm now trying to create BLIS static library for a target microcontroller (NXP MPC5748G, having 32bit PowerPC architecture on which code built by GCC is mounted) through the corresponding cross compiler with its specialized C standard library called EWL, but I'm in trouble while building the code. 

Prerequisite: 
- having the corresponding GCC cross-compiler (powerpc-eabivle-gcc) for the target device on the host machine in the directory of `${HOME}/NXP/S32DS_Power_v2.1/S32DS/build_tools/powerpc-eabivle-4_9/bin/'
- having the corresponding standard C libraries (EWL) for the target device on the host machine in the directory of `${HOME}/NXP/S32DS_Power_v2.1/S32DS/build_tools/e200_ewl2/'

What I tried was:
- Go into `config/generic/` and populate `make_defs.mk` with details / flags on the target cross compiler. (I attach the `make_defs.mk` file I used.)
- Run command `./configure CC=${HOME}/NXP/S32DS_Power_v2.1/S32DS/build_tools/powerpc-eabivle-4_9/bin/powerpc-eabivle-gcc --disable-shared --disable-threading --disable-system --enable-verbose-make generic` to genrate Makefile.
- As `make -j` encounters "Cannot determine operating system" error, I go into `include/generic/blis.h` file, comment two blocks causing the error, and explicitly define `#define BLIS_OS_GNU 1` (I attach the modified header file I used.)
- Run `make -j`, but it shows warning statements pertaining to using `fprintf`. (I attach the build logs (containing standard statements and warnings separately) as well.)

Anyway, it ends up creating the static library file `libblis.a`. But when I try to link it with the BLIS tutorial code, it throws similar errors as mentioned above; this is not so surprising as the library file has been built 'incomplete'. 

As the `system root` (where the standard C libraries are located) has been set properly to be `${HOME}/NXP/S32DS_Power_v2.1/S32DS/build_tools/e200_ewl2/' and `stdio` is in there, I believe that it wouldn't cause such issues but it didn't. This problem has eaten up to much time and effort so far. It would be really appreciated if anyone knows a silver bullet to get around such an issue. Thanks in advance!

##### files mentioned above #####
[files.zip](https://github.com/flame/blis/files/7015154/files.zip)

hominhquan:
[...]

PR #546, opened 2021-09-20
blis.h: re-enable 8e0c425 and reorder include order to compile non-system mode

hominhquan:
- Before, OS-detection is always done even with --disable-system, because
bli_system.h is included before bli_config.h. This may explain why
8e0c425 has been reverted in 2be78fc.
- This commit swaps their order (/!\\) : bli_config.h before bli_system.h.
- Also better support future non-system platforms (#532)
```
// NOTE: bli_config.h header must be included before any BLIS header.
// It is bootstrapped by ./configure and does not depend on later
// headers. Moreover, these configuration variables are necessary to change
// some default behaviors (e.g. disable OS-detection in bli_system.h in case
// of --disable-system).
```

### DIFF
--- a/frame/include/bli_system.h
+++ b/frame/include/bli_system.h
@@ -70,7 +70,7 @@
 #endif
 
 // Determine the target operating system.
-//#if defined(BLIS_ENABLE_SYSTEM)
+#if defined(BLIS_ENABLE_SYSTEM)
   #if defined(_WIN32) || defined(__CYGWIN__)
     #define BLIS_OS_WINDOWS 1
   #elif defined(__gnu_hurd__)
@@ -94,9 +94,9 @@
   #else
     #error "Cannot determine operating system"
   #endif
-//#else // #if defined(BLIS_DISABLE_SYSTEM)
-//  #define BLIS_OS_NONE
-//#endif
+#else // #if defined(BLIS_DISABLE_SYSTEM)
+  #define BLIS_OS_NONE
+#endif
 
 // A few changes that may be necessary in Windows environments.
 #if BLIS_OS_WINDOWS
--- a/frame/include/blis.h
+++ b/frame/include/blis.h
@@ -48,6 +48,15 @@ extern "C" {
 // NOTE: PLEASE DON'T CHANGE THE ORDER IN WHICH HEADERS ARE INCLUDED UNLESS
 // YOU ARE SURE THAT IT DOESN'T BREAK INTER-HEADER MACRO DEPENDENCIES.
 
+// -- configure definitions --
+
+// NOTE: bli_config.h header must be included before any BLIS header.
+// It is bootstrapped by ./configure and does not depend on later
+// headers. Moreover, these configuration variables are necessary to change
+// some default behaviors (e.g. disable OS-detection in bli_system.h in case
+// of --disable-system).
+#include "bli_config.h"
+
 // -- System and language-related headers --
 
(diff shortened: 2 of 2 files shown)


hominhquan:
Hummm, it seems to be broken on windows
```
./frame/thread\bli_pthread.h:77:5: error: unknown type name 'HANDLE'
    HANDLE handle;
    ^
```

hominhquan:
Looks like there is compilation failure of some configure-time C files which are including BLIS header but without `bli_config.h`:
- `./configure`: `auto_detect()`
- `build/detect/config/config_detect.c`

fgvanzee:
> Hummm, it seems to be broken on windows
> 
> ```
> ./frame/thread\bli_pthread.h:77:5: error: unknown type name 'HANDLE'
>     HANDLE handle;
>     ^
> ```

Yes, this is the issue that I encountered as well. I had absolutely *no* idea at the time what could have possibly caused this, hence why I reverted 8e0c425 (in 2be78fc). But I think I now have some understanding. There's something about the Windows builds on AppVeyor that is causing `BLIS_ENABLE_SYSTEM` to be undefined. Specifically, I'm guessing the problem is in the bit of code that is executed for configure-time hardware detection.

hominhquan:
@fgvanzee so we have the same investigation result. Now since I'm not familiar to that part of code, can you tell if including `bli_config.h` in this code solves the compilation issue ? (and if yes, so why `bli_config.h` has not been included before ?)

fgvanzee:
I don't think we want to `#include bli_config.h` in `config_detect.c`. The idea is we want that source file to depend on as few headers as possible.

I think I see two possible solutions.
1. Insert `#define BLIS_ENABLE_SYSTEM` before line 38 of `config_detect.c`. This will cause the correct cpp branch to be processed in `bli_system.h`.
2. Change the structure of the cpp conditional in `bli_system.h` so that if neither `BLIS_ENABLE_SYSTEM` nor `BLIS_DISABLE_SYSTEM` is defined, it executes the `BLIS_ENABLE_SYSTEM` branch.

I am partial to (1) because (2) is kinda subtle, and I prefer to avoid code behavior that depends on something subtle.

hominhquan:
Yes, I vote for (1), too

fgvanzee:
> I am partial to a more aggressive version of (2): ONLY ever define BLIS_DISABLE_SYSTEM and have BLIS_ENABLE_SYSTEM be the default. It makes no sense to have two macros which are always the opposite of each other.

So it sounds like you take issue with how most of `bli_config.h` works.

devinamatthews:
I guess, so please ignore me.

hominhquan:
In fact, why don't we check for BLIS_ENABLE_SYSTEM just before yielding the error "Cannot determine system", because that `#error` is the root of all, am I right ? 
```
  #else
    #if BLIS_ENABLE_SYSTEM
      #error "Cannot determine operating system"
    #else
      #warning "Hey, you are compiling with --disable-system mode"
    #endif  // BLIS_ENABLE_SYSTEM
  #endif
```

fgvanzee:
Fair suggestion. But my interpretation of the `--disable-system` option is that you want to be able to use it even if an OS is actually present, in which case we would want to have the `#if defined(BLIS_ENABLE_SYSTEM)` at the beginning.

fgvanzee:
I think this PR is now mooted by fb93d24. @hominhquan, can you confirm?

hominhquan:
@fgvanzee yes. Thanks for the commit. I guess we can close close this now.

PR #547, opened 2021-09-21
Make reference ukernels MR/NR sizes configurable

hominhquan:
- Those reference micro-sizes are hardcoded for some performance purposes
(#pragma omp simd). They may work most of the time, except when user begins
tuning the MR/NR sizes.
- This commit exposes them as configurable by user in bli_family_<arch>.h,
makes generation of reference ukernels more robust/consistent. Some explanation
is added to put future developers on guard (should be added in wiki as well).

### DIFF
--- a/docs/ConfigurationHowTo.md
+++ b/docs/ConfigurationHowTo.md
@@ -239,6 +239,29 @@ The value `BLIS_HEAP_STRIDE_ALIGN_SIZE` defines the alignment used for so-called
 
 The value `BLIS_POOL_ADDR_ALIGN_SIZE` defines the alignment used when allocating blocks to the memory pools used to manage internal packing buffers. Any block of memory returned by the memory allocator is guaranteed to be aligned to this value. Aligning these blocks to the virtual memory page size (usually 4096 bytes) is standard practice.
 
+_**Reference micro-kernel sizes.**_ Using pre-implemented reference micro-kernels could be useful to port and get BLIS running quickly on a new architecture, without spending time implementing a set of optimized kernels, yet.
+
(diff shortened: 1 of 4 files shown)


hominhquan:
Update with added doc in `docs/ConfigurationHowTo.md` (113a1a9)

devinamatthews:
@hominhquan I'm planning to address this issue within a broader update of the microkernel layer over the next few months. Is having a fix for this issue being merged into `master` a priority for you?

hominhquan:
@devinamatthews Yes this fix is useful to me. It may also prevent other people from falling to the issue, at least in the next few months. So yes I would like to see it merged into master.

devinamatthews:
@fgvanzee what do you think? The other fix of course is to disable the vectorized reference kernel.

fgvanzee:
> @fgvanzee what do you think? The other fix of course is to disable the vectorized reference kernel.

Is disabling the vectorized reference gemm ukernel, as you propose, really just a matter of changing `#if 1` to `#if 0` on line 37 of `ref_kernels/3/bli_gemm_ref.c`? If so, and *if* that fix works for @hominhquan, then I'm somewhat partial to that solution. But if @hominhquan needs the individual MR/NR macro constants for other purposes beyond disabling the vectorization, then I suppose that fix would not work.

devinamatthews:
@fgvanzee I believe it is that easy.

hominhquan:
@devinamatthews  or @fgvanzee Can you tell if this has been addressed in any ongoing development and should I drop this PR ?

fgvanzee:
@devinamatthews might be able to comment further, but yes, this is part of ongoing development. We've spoken about Devin's idea as it pertains to this topic and I think it will work for your purposes, too, @hominhquan.

devinamatthews:
@hominhquan after some pending PRs I'm planning to work on overhauling the way kernels are handled. As part of that, I plan to reintroduce the MR/NR macros so that reference kernels can use them to do compile-time optimizations safely.

hominhquan:
Ok, so I drop this PR and wait for your coming update.

PR #549, opened 2021-09-28
Enable user-customization of packing

devinamatthews:
This PR adds functionality to override the "default" packing microkernel (which packs one MR*K or NR*K micropanel) and variant (kernel, which packs an entire matrix block/panel). A void* field on obj_t allows users to pass arbitrary additional parameters to packing microkernels and variants.


devinamatthews:
@hominhquan yes, things got reordered and so we ended up with a race condition. I think now it is pretty compartmentalized and modular so that a custom packing kernel could use `bli_packm_init()` without taking any extra precautions.

devinamatthews:
Finally. @fgvanzee this is ready to review when you're done with #531.

devinamatthews:
@fgvanzee ready to review

fgvanzee:
I think we're done. I'll compile a monolithic commit message and then squash.

PR #551, opened 2021-09-28
Add `-mincoming-stack-boundary` to `haswell` kernel flags.

### LINKED ISSUE #550: Force stack alignment for AVX kernels
The perfect storm happens:

1. On macOS
2. With gcc (in my case 11.2)
3. With debugging enabled (`CFLAGS="-g -O0"`)
4. Using the `haswell` configuration

What occurs in my case is that `bli_saxpyv_zen_int10` is compiled with flags for AVX, and GCC makes the assumption that the incoming stack pointer is 32-byte aligned. Then, because of the debug flags, horrible code is emitted which spills vector registers (`__m256`) to the stack. This spilling uses aligned move instructions which require 32-byte alignment. GCC dutifully 32-byte aligns all offsets from the stack pointer to satisfy this constraint. Here is the catch: if the incoming stack point is *NOT* actually 32-byte aligned (as it isn't in BLIS, the calling code is not compiler for AVX but for general x86_64), then we encounter a segfault.

It has proven quite difficult to actually force GCC to align the stack using e.g. `andq $-32, %rsp`. The only thing that seems to work is `-mincoming-stack-boundary=3` (*not* 4!). I will add this to the default `haswell` flags since so far I can only confirm the problem on a mac. I'll see about Linux shortly.

devinamatthews:
This fixes #550.

### DIFF
--- a/config/haswell/make_defs.mk
+++ b/config/haswell/make_defs.mk
@@ -1,6 +1,6 @@
 #
 #
-#  BLIS    
+#  BLIS
 #  An object-based framework for developing high-performance BLAS-like
 #  libraries.
 #
@@ -65,10 +65,10 @@ endif
 # they make explicit use of the rbp register.
 CKOPTFLAGS     := $(COPTFLAGS) -O3 -fomit-frame-pointer
 ifeq ($(CC_VENDOR),gcc)
-CKVECFLAGS     := -mavx2 -mfma -mfpmath=sse -march=haswell
+CKVECFLAGS     := -mavx2 -mfma -mfpmath=sse -march=haswell -mincoming-stack-boundary=3
 ifeq ($(GCC_OT_4_9_0),yes)
 # If gcc is older than 4.9.0, we must use a different label for -march.
-CKVECFLAGS     := -mavx2 -mfma -mfpmath=sse -march=core-avx2
+CKVECFLAGS     := -mavx2 -mfma -mfpmath=sse -march=core-avx2 -mincoming-stack-boundary=3
 endif
 else
 ifeq ($(CC_VENDOR),icc)


devinamatthews:
@fgvanzee this is to fix the bug we talked about. Finally found a flag that actually does what I want.

fgvanzee:
Brilliant. Do we need to modify any other subconfigs' makefile fragments? Or do we have reason to believe this is a `haswell`-only thing?

devinamatthews:
It is a mac-only thing so essentially it is `haswell`-only.

devinamatthews:
@fgvanzee  Apparently `-mincoming-stack-boundary=3` counteracts `-fomit-frame-pointer` and so the gemmsup kernels don't comple because they use `rbp`... now, I do have a patch which removes `rbp` from those kernels, shall I add that to the PR?

fgvanzee:
> so the gemmsup kernels don't comple because they use `rbp`... now, I do have a patch which removes `rbp` from those kernels, shall I add that to the PR?

I'm kinda curious how you removed the use of `rbp`. I remember maxing out the use of GPRs and having no room to spare.

(AMD has lots of sup-like kernels that they probably programmed in a similar style that they will have to update too, someday, if we go through with this.)

Whether you do it here or in another PR -- either way is fine. Although I guess it makes a certain sense to include it here since it's linked to the `-mincoming-stack-boundary` option.

devinamatthews:
I think there is little hope actually. Even removing usage of `rbp` I get the dreaded "impossible register contraints". Since this only affects a very very limited use case (more limited than I initially thought), then I think it may have to go unfixed.

fgvanzee:
> I think there is little hope actually. Even removing usage of `rbp` I get the dreaded "impossible register contraints". Since this only affects a very very limited use case (more limited than I initially thought), then I think it may have to go unfixed.

That's lame. Thanks for your valiant efforts, though. Nobody can say we didn't try to bend over backwards to thread this needle.

PR #553, opened 2021-10-04
Add an option to use an @rpath-dependent install_name on macOS

devinamatthews:
This PR does the following:

1. Adds an option `--enable-rpath/--disable--rpath` (default disabled) to use an install_name starting with @rpath/. Otherwise, set the install_name to the absolute path of the install library, which was the previous behavior. The new behavior is more like Linux where the library is relocatable but users must specify an RPATH entry in their binaries or set LD_LIBRARY_PATH/DYLD_LIBRARY_PATH.
2. Fixes an issue where the testsuite binaries would not run on macOS with `--disable-static` before installation. E.g. now one can do "make test" or "make check" with `--disable-static --enable-rpath` on macOS before or without installing.
3. Makes the testsuite binaries relocatable (as part of the build tree). Might be helpful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -725,19 +725,13 @@ else
 	@$(RANLIB) $@
 endif
 
-# first argument: the base name of the BLAS test driver.
-define make-blat-rule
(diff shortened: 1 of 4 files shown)


devinamatthews:
Apologies if there is a user named @rpath...

PR #559, opened 2021-10-12
bli_pool: fix potential insufficient pool-growing

hominhquan:
- The current growing mechanism doubles the length of the block_ptrs array
everytime the array length needs to be re-adjusted, but did not take in account
the new total number of blocks, that in some cases, doubling the block_ptrs
array length is not enough.
- Two illustrating examples are given in the code.
- This commit fixes it by taking max with num_blocks_new, too.
- This commit also fixes incidentally the memory corruption of the growing
process of any empty-, hence ill-initialized pool (block_ptrs array length = 0).

### DIFF
--- a/frame/base/bli_pool.c
+++ b/frame/base/bli_pool.c
@@ -373,7 +373,15 @@ void bli_pool_grow
 	{
 		// To prevent this from happening often, we double the current
 		// length of the block_ptrs array.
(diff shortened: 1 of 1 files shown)


fgvanzee:
Yep, that's definitely a bug. Great work on tracking this down, @hominhquan.

Issue #437, opened 2020-08-12
Mis-balanced work between JR/IR threads in edge-macro-blocks


hominhquan:
Let call  `macro-block` an MC-by-NC block, which is traditionally processed by a call to `macro-kernel`
Let call `edge-macro-block` either an ME-by-NC or an MC-by-NE block, where `0 < ME < MC` and `0 < NE < NC` (`E` stands for edge). In short, edge-macro-blocks occur when the C matrix size (M or N) is not multiple of MC (or NC). 

I observed the mis-balancing in my old version of BLIS on edge-macro-blocks . After checking with the latest version of BLIS, even with the slab or round-robin dispatch, I think the issue still remains when all following conditions are met:
1. Edge-macro-block occurs, where `n_iter` (or `m_iter`) in the macro-kernel will be smaller than the normal `NC/NR` (or `MC/MR`).
2. User set `BLIS_JR_NT > 1`, for example  `BLIS_JR_NT = NC/NR` (one n_iter per jr_thread) or  `BLIS_JR_NT = NC/(2*NR)` (two n_iter per jr_thread). Idem for `BLIS_IR_NT > 1`, for example  `BLIS_IR_NT = MC/MR` (one m_iter per ir_thread) and so on.
3. User tuned all `BLIS_*_NT` so that their product is equal to the number of physical cores.

If these conditions are true, we can have for example an edge-macro-block with `n_iter` smaller than the user-defined `BLIS_JR_NT`. Since all those threads are spawned, the "trailing" JR threads will standby looking at a few other JR threads working on edge-n_iter blocks. The situation is also the same for the IR-loop. And the slab/rr dispatch can not help in this case, since the work is dispatched **_separately_** between JR-loop and IR-loop. As a result, we can have up to half of cores (or even worse) not doing any computation.

One solution I can see is to detect this edge case and to dynamically "fusion" the two loops into a flatten 1D workspace of `c11` micro-blocks and dispatch them equally on `jr_nt * ir_nt`  threads

For instance, I did this on the gemm macro-kernel in my BLIS version 0.2.2 and it fixes the mis-balancing issue:

```lang=c
	/* fused num_threads in JR and IR */ 
	dim_t num_threads = jr_num_threads * ir_num_threads; 
	
	/* fused number of c11 micro-blocks */
	dim_t num_blocks = m_iter * n_iter; 
	dim_t num_blocks_per_thread = num_blocks / num_threads; 

	/* possible trailing c11 blocks */
[...]

PR #562, opened 2021-10-20
Detect and deal with mis-balancing in GEMM macro-kernel (#437)

hominhquan:
Details:
- In some multi-threading schemes, JR_NT and IR_NT may produce idle threads
not performing any computation.
- This commits detect such situation and implement a collapse of JR/IR loops.

Note: run `git show -w <sha1>` for more readability, due to new indent level by `if()`

### DIFF
--- a/frame/3/gemm/bli_gemm_ker_var2.c
+++ b/frame/3/gemm/bli_gemm_ker_var2.c
@@ -312,55 +312,151 @@ void PASTEMAC(ch,varname) \
 	dim_t ir_nt  = bli_thread_n_way( caucus ); \
 	dim_t ir_tid = bli_thread_work_id( caucus ); \
 \
-	dim_t jr_start, jr_end; \
-	dim_t ir_start, ir_end; \
-	dim_t jr_inc,   ir_inc; \
+	/* Mis-balancing detection: if n_iter (or m_iter) is not multiple of
+	  jr_nt (or ir_nt), then we collapse the two JR/IR loops and dispatch
+	  work on jr_nt * ir_nt threads. */ \
+	const bool misbalancing = ((n_iter % jr_nt) != 0) || ((m_iter % ir_nt) != 0); \
 \
-	/* Determine the thread range and increment for the 2nd and 1st loops.
-	   NOTE: The definition of bli_thread_range_jrir() will depend on whether
-	   slab or round-robin partitioning was requested at configure-time. */ \
-	bli_thread_range_jrir( thread, n_iter, 1, FALSE, &jr_start, &jr_end, &jr_inc ); \
-	bli_thread_range_jrir( caucus, m_iter, 1, FALSE, &ir_start, &ir_end, &ir_inc ); \
-\
-	/* Loop over the n dimension (NR columns at a time). */ \
-	for ( j = jr_start; j < jr_end; j += jr_inc ) \
+	if ( !misbalancing ) \
 	{ \
-		ctype* restrict a1; \
-		ctype* restrict c11; \
-		ctype* restrict b2; \
+		/* Use traditional two loops */ \
+		dim_t jr_start, jr_end; \
+		dim_t ir_start, ir_end; \
+		dim_t jr_inc,   ir_inc; \
 \
-		b1 = b_cast + j * cstep_b; \
-		c1 = c_cast + j * cstep_c; \
+		/* Determine the thread range and increment for the 2nd and 1st loops.
+		   NOTE: The definition of bli_thread_range_jrir() will depend on whether
+		   slab or round-robin partitioning was requested at configure-time. */ \
+		bli_thread_range_jrir( thread, n_iter, 1, FALSE, &jr_start, &jr_end, &jr_inc ); \
+		bli_thread_range_jrir( caucus, m_iter, 1, FALSE, &ir_start, &ir_end, &ir_inc ); \
 \
-		n_cur = ( bli_is_not_edge_f( j, n_iter, n_left ) ? NR : n_left ); \
-\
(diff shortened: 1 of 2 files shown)


fgvanzee:
@devinamatthews Curious: did you ever take a look at this PR? What are your thoughts?

devinamatthews:
@fgvanzee I'll look at the changes in more detail. This was dicussed in #437 and only affects cases where `BLIS_JR_NT` is probably too large.

devinamatthews:
Yeah, I'm not sure I'm comfortable with this. I really think the better answer is just to not use so many threads in the JR loop. @hominhquan  Is there a particular reason to set `BLIS_JR_NT` so high?

hominhquan:
> @hominhquan Is there a particular reason to set `BLIS_JR_NT` so high?

Yes, on our MPPA processor, on which matrix tiles (A, B, C) are copied to scratchpad by DMA, memory-footprint is a big concern. We have 16 cores on each "cluster", but only spawn **one** simultaneous NC-loop (aka `BLIS_JC_NT = 1`) on the top, then going down and sharing the computation between the cores in the macro-kernel. So the macro-kernel dispatch is subjected to   `BLIS_JR_NT * BLIS_IR_NT = 16`. 

Spawning two NC-flows (`BLIS_JC_NT = 2`) can lighten the above constraint, but will inevitably double the local DMA scratchpad requirement, that we can't afford. (see illustration https://user-images.githubusercontent.com/1337056/178685971-7510aca3-ab44-46b2-aed3-43c25985f67b.png)

Talking more generally, I think collapsing the two JR/IR loops in the macro kernel could help reducing the loop overhead and give better load-balancing in any threading scheme, and so more "hardware-friendly". I remember (I may be wrong) having seen some commits or discussions about adding custom hand-optimized macro-kernels into BLIS ? If yes, I am curious to know the motivation behind that.

Note: In this PR I kept the original two-nested-loops and added a new collapsed one, which sadly made bigger code, but if it was my personal repo, I would keep only the collapsed version.

devinamatthews:
You aren't doing any threading along the M dimension (`BLIS_IC_NT`)?

hominhquan:
Setting IR_NT=4 and JR_NT=4 could improve the situation , but is not the ultimate solution. Here is an example of how the two-nested-loop dispatch can be mis-balanced on edge-blocks:

M = N = 3000, MC = NC = 256, MR = 8, NR = 16.
Possible edge-macro-block is (3000 % 256): 184-by-184 => m_iter = 23, n_iter = 12 (ceil(11.5))

- If  BLIS_JR_NT = 16, BLIS_IR_NT = 1, there will be only 12 threads working on the JR-loop, and 4 threads in standby.
- If  BLIS_JR_NT = 8, BLIS_IR_NT = 2, there will be enough work for the 8 JR-threads for the first batch (8 < n_iter), 
but the second batch there will be only 4 JR-threads working (n_iter - 8 = 4), and 4 JR-threads in standby.
Idem for IR-threads (since m_iter = 23 not multiple of BLIS_IR_NT)
- If  BLIS_JR_NT = 4, BLIS_IR_NT = 4, the dispatch will be perfect in JR-loop (n_iter = 12 multiple of BLIS_JR_NT), but not in IR-loop where m_iter = 23 not multiple of BLIS_IR_NT. There will be 3 IR-threads working, and 1 IR-thread idle x 4 JR-threads = 4 threads in standby.

In this case, setting IR_NT=4 and JR_NT=4 removes the misbalancing for edge-blocks of size 256x184 or 184x256 (right and bottom), but not the final edge-block 184x184 in the right-bottom, even that it impacts only one block of 184x184.

hominhquan:
> Is this also a memory thing? Parallelizing along the IC loop would definitely be preferable.
Yes, still the memory-footprint stuff. Setting IC_NT to something bigger than 1 will also increase the DMA-scratchpad requirement.

fgvanzee:
@devinamatthews For context: The only reason I asked about this PR when I did is because an industry partner asked about it during a conference call, so I wanted to be able to tell them what was going on with it. Now that I've looked at it more closely, I can say that I'm not comfortable with the current implementation. But if we can find a more elegant way of expressing the logic that doesn't involve so much code duplication, I'm open to considering it.

> @fgvanzee what might happen if the collapsed version were used all the time?

I would need to study the code more before answering this.

hominhquan:
> > But if we can find a more elegant way of expressing the logic that doesn't involve so much code duplication, I'm open to considering it.
> 
> This was my concern as well, and hence the question.

May be I was way too careful when doing this PR and kept the original implementation. As I said before, if this was my code I would replace the two-nested loops by the collapsed one.

hominhquan:
> But if we can find a more elegant way of expressing the logic that doesn't involve so much code duplication, I'm open to considering it.
>> @fgvanzee what might happen if the collapsed version were used all the time?
> I would need to study the code more before answering this.

@devinamatthews @fgvanzee I've looked at the code again, and saw that `thrinfo_t *thread` and `thrinfo_t* caucus` were already created in the thread-info tree. My modif of using still another temporary `thrinfo_t` is not optimal either.

Further, I would like to know your thoughts on the relevance of always collapsing the JR/IR loops. Is there any other macro-kernel  in which collapsing JR/IR loops is bad or difficult or impossible ? (1m ? trsm ? mixed-precision ? mixed-domain ?)
- If the answer is yes, then we can apply the collapsing case-by-case, where the modification is trivial (e.g. real-real gemm)
- If the answer is no, and if you are convinced that collapsing is good for performance, we might apply it everywhere, and reconsider the utility of "caucus" in the thread-info tree.

hominhquan:
Closed as implemented in #695.

PR #565, opened 2021-10-22
bli_pool: add sanity check in bli_pool_free_block()

hominhquan:
Details:
- In case of pool reinit due to a thread asking for a bigger block, the
  pool is re-initialized and all blocks are free-ed, including potential
  ones already checked out and being used elsewhere. Those empty slots
  are marked with NULL pointer and hence we should skip calling free()
  on them (with subtraction by offset_size).
- The orphaned blocks and their pointers will be free-ed later when
  checking in back by the existing check-in mechanism.


devinamatthews:
@hominhquan I believe the problems leading to this PR were fixed recently in #670 and #671. If not, let me know.

PR #566, opened 2021-10-26
Added BLAS, CBLAS interfaces and test functions for gemm_batch and axpbyv APIs.

Meghana-vankadari:
Added BLAS, CBLAS interfaces and test functions for gemm_batch and axpbyv APIs.

AMD BLIS Upstream:

This PR includes following commits for AMD BLIS version 3.0.1

f001b189  Added few changes in test_axpbyv.c file
3440b84b Added BLAS and CBLAS interfaces for axpby API
9869388c Added CBLAS interface and test file for gemm_batch API
9b4be5fe Added BLAS interface for gemm_batch API


fgvanzee:
@Meghana-vankadari @devinamatthews I'm having second thoughts about the templatization for `axpby` and `gemm_batch` CBLAS APIs. The reason is that in order to follow CBLAS's calling conventions, the real domain matrices and vectors are passed via typed pointers (e.g. `const float*` and `const double*`), but for the complex domain interfaces, matrices and vectors are passed via `const void*` pointers. This can't easily be encoded into a cpp macro template.

So now I'm thinking that it may be better to leave the code as-is.

devinamatthews:
Well, it can be done with macros fairly easily but it doesn't really matter much either way. I am fine with non-templated code I guess as long as it is not our responsibility to maintain it.

fgvanzee:
> I am fine with non-templated code I guess as long as it is not our responsibility to maintain it.

Agreed.

PR #578, opened 2021-11-16
Test driver fix for her, her2, herk and her2k

madanm3:
Bug fix done driver code. 
- API mismatch were fixed and data type supported test has been added.


fgvanzee:
@madanm3 Thanks for your fixes. I added the real domain `syr`/`syr2`/`syrk`/`syr2k` calls to the driver files since the driver files are not literally intended to refer only to complex domain `her`/`her2`/`herk`/`her2k`. Rather, we often use the `herk` operation, for example, to refer to the pair of operations including real domain `syrk` *plus* complex domain `herk`. (Notice that if you perform `herk` on a real domain matrix, you've actually performed the equivalent of `syrk` on the same matrix.)

fgvanzee:
PS: I cancelled the Travis CI build since there is no way these changes could affect the library.

PR #590, opened 2021-12-23
Added BLAS Extension API's: CBLAS_?GEMM3M

BhaskarNallani:
Induced M1 method is enabled for CGEMM3M and ZGEMM3M


frame/3/bli_l3_ind.c
```diff
@@ -58,7 +58,7 @@ bool bli_l3_ind_oper_st[BLIS_NUM_IND_METHODS][BLIS_NUM_LEVEL3_OPS][2] =
         /*   gemm           gemmt          hemm           herk           her2k          symm
              syrk           syr2k          trmm3          trmm           trsm  */
         /*    c     z    */
-/* 1m   */ { {FALSE,FALSE}, {FALSE,FALSE}, {FALSE,FALSE}, {FALSE,FALSE}, {FALSE,FALSE}, {FALSE,FALSE},
+/* 1m   */ { {TRUE,TRUE}, {FALSE,FALSE}, {FALSE,FALSE}, {FALSE,FALSE}, {FALSE,FALSE}, {FALSE,FALSE},
```
fgvanzee, line 61: @BhaskarNallani It appears that you are enabling 1m by default for `gemm`, which will cause it to have precedence over a native microkernel (if it exists). Can you elaborate on why this change was made / needed?

If your goal is for `gemm3m` to use the 1m implementation of `gemm`, I think we will need to find another way to make this happen.

fgvanzee: FYI: I have found a more elegant solution to this, which I am in the process of implementing.

fgvanzee:
@BhaskarNallani Please take a look at this updated branch and let me know if it works for you.

BhaskarNallani (commented):
If 1m is disabled then you want to solve using native path for gemm3m?

fgvanzee:
@BhaskarNallani My solution was to ignore whether 1m is enabled or disabled in `bli_l3_ind.c` (which is considered a user API structure), and unconditionally pass a 1m context into `bli_gemm_front()`. 

In other words, we silently implement `gemm3m` in terms of 1m (for now) regardless of whether the user wants `gemm` via 1m or native execution (which I consider a separate issue).

See `frame/compat/extra/bla_gemm3m.c` for details.

fgvanzee:
@BhaskarNallani Let me know if you have any questions about this PR during our upcoming meeting.

BhaskarNallani:
@fgvanzee , I agree with the changes you have done to eliminate enabling 1m method in table and calling directly GEMM3M api function by disabling sup. Please merge the changes.

PR #571, opened 2021-10-29
Added opt kernels for axpyf, copyv, swapv

Meghana-vankadari:
AMD BLIS Upstream:

This PR includes following commits for AMD BLIS version 3.0.1

62ff1a59 Optimized axpyf kernel for scomplex datatype
a569d6cf Added opt kernels for SWAPV
911166de Adding a simd kernel for copyv function
297aa226 Added 16x4 AXPYF kernel for zen2 config

PR #597, opened 2022-01-11
Framework cleanup and addition of opt kernels for axpyf, copyv, swapv

dzambare:
This PR replaces PR#571

The framework code is cleaned up to remove AMD specific code changes from common files. 

However, to get the maximum performance we have to skip framework and call kernels directly from cblas/blas layer, because of this there are some framework files where AMD specific code is needed. For these files exiting files are duplicated and (_amd) prefix is added in the file name. The build system will pick AMD specific files if the build is for any of the AMD architecture. 

The AMD specific changes can work on other platforms as well however it is more of architectural decision that needs to be taken by BLIS maintainers if it it can be included as a normal flow for all architectures.  

@Meghana-vankadari, Kindly verify that all changes from PR #571 are taken care.


Makefile
```diff
@@ -1343,3 +1365,9 @@ else
 	@- $(RM_F) $@
 endif
 
+debug:
```
dzambare, line 1368: This is not needed in final commit.

frame/compat/cblas/src/cblas_dcopy.c
```diff
@@ -7,7 +7,10 @@
  *
  * Written by Keita Teranishi.  2/11/1998
  *
+ * Copyright (C) 2020, Advanced Micro Devices, Inc.
```
dzambare, line 10: Checkout this file from flame/master.

frame/compat/cblas/src/cblas_dcopy.c
```diff
@@ -20,6 +23,7 @@ void cblas_dcopy( f77_int N, const double *X,
    #define F77_incX incX
    #define F77_incY incY
 #endif
+
```
dzambare, line 26: Checkout this file from flame/master.

frame/compat/cblas/src/cblas_scopy.c
```diff
@@ -7,6 +7,8 @@
  *
  * Written by Keita Teranishi.  2/11/1998
  *
+ * Copyright (C) 2020, Advanced Micro Devices, Inc.
```
dzambare, line 10: Checkout this file form flame/master.

test/Makefile
```diff
@@ -224,4 +224,4 @@ clean: cleanx
 
 cleanx:
 	- $(RM_F) *.o *.x
-
```
dzambare, line 227: No real change here, remove this file from the changeset.

dzambare:
@fgvanzee : We were planning to move all AMD specific files inside a subfolder, can it be accommodated now? otherwise these changes seems fine to me.

frame/compat/bla_copy_amd.c
```diff
@@ -0,0 +1,146 @@
+/*
+
+   BLIS
+   An object-based framework for developing high-performance BLAS-like
+   libraries.
+
+   Copyright (C) 2014, The University of Texas at Austin
```
dzambare, line 7: Copyright may need to be updated on newly added files.

fgvanzee: I've updated the copyrights.

As an aside, I've noticed in the past that AMD has been including "All rights reserved" on the copyright notice in some files. My understanding is that this language is unnecessary because it is implied by the copyright, and therefore is redundant [1].

[1] https://www.legalzoom.com/articles/the-term-all-rights-reserved-explained

fgvanzee:
> @fgvanzee : We were planning to move all AMD specific files inside a subfolder, can it be accommodated now? otherwise these changes seems fine to me.

Done.

fgvanzee:
@dzambare Note that the way the `_amd.c` files are chosen to be compiled has changed. Previously, if any of the `zen[23]` or `amd64` configurations is targeted, the files were used instead of their generic counterparts. Now, we have a `configure` option, `--enable-amd-frame-tweaks`, that is a *request* to do so. When requested, a subsequent check is then performed to see if the targeted configuration is actually eligible (i.e., `zen[23]` or `amd64`). The option defaults to "disabled."

In AMD's fork of BLIS, you may wish to change the default so that it is enabled by default. This can be done with a one-line change in `configure` to line 2471:
```bash
# Change initial value to 'yes' to change the default behavior. Don't forget
# to update the './configure --help' text to reflect the new default so that
# users do not get confused. 
enable_amd_frame_tweaks='no'
```

frame/compat/bla_copy_amd.c
dzambare: @fgvanzee , these are the specific guidelines provided by AMD legal team.

dzambare:
Thanks @fgvanzee , the solution with extra configuration flag looks good, really appreciate your help in getting these change integrated.

PR #600, opened 2022-01-17
fix stale code in bli_thrcomm_barrier - use correct struct member names.

jeffhammond:
code was disabled and not properly maintained.  this updates the struct member names to be correct.

Signed-off-by: Jeff Hammond <jehammond@nvidia.com>


devinamatthews:
@jeffhammond this code should be disabled since it is superseded by the call to `bli_thrcomm_barrier_atomic`.

jeffhammond:
Ok. Then it should be deleted.

devinamatthews:
Field is particularly fond of `#if 0`...

Issue #604, opened 2022-01-28
thread barriers need backoff


jeffhammond:
This code leads to serious problems when hardware threads are oversubscribed.  Adding `sched_yield()` reduces the problem by one or two orders of magnitude by triggering the kernel to swap threads so progress happens more quickly.

```c
// If the current thread is NOT the last thread to have arrived, then
// it spins on the sense variable until that sense variable changes at
// which time these threads will exit the barrier.
while ( __atomic_load_n( &comm->barrier_sense, __ATOMIC_ACQUIRE ) == orig_sense )
     ; // Empty loop body.
```

I tried no-op instructions but those do not help in the oversubscribed case, because they don't trigger a context switch.  Those backoffs are appropriate when memory access contention is the issue.

This is related but complementary to https://github.com/flame/blis/issues/603.  This is a new version of https://github.com/flame/blis/pull/82.

My proposed fix will allow a user to disable `sched_yield()` but I assert we need it enabled in the distribution builds of BLIS because quality-of-service is more important than the last bit of performance in the general case.  Benchmarking use cases can disable it if it is expected to matter there.

# References

* https://github.com/flame/blis/issues/588#issuecomment-1014416925
* https://github.com/flame/blis/issues/588#issuecomment-1021624825

devinamatthews:
@jeffhammond `sched_yield` is too heavyweight and not portable enough to be used all the time. I will update #82 with a general framework for config-specific behavior and then we can start filling in the actual implementation.

devinamatthews:
@jeffhammond suggestions for any specific architectures?

chillenb:
Libgomp barrier uses futex for linux. 

https://github.com/gcc-mirror/gcc/blob/master/libgomp/config/linux/bar.c

Would that be appropriate? Maybe not portable enough?

jeffhammond:
1. We cannot use GOMP because of the license.
3. GOMP barrier performance is worst-in-class for OpenMP so this is not an implementation to which anyone should aspire.  The only scenario where it is good is oversubscription, which isn't an HPC use case.

devinamatthews:
futex is reasonable for *after* you have backed off a few times.

chillenb:
Thanks for the information. It is useful to know that about GOMP; apologies for reminding you of any unpleasant GOMP-associated memories.

I tried to see what LLVM/Intel OpenMP is doing. For your entertainment, here are some links to sections of code that look relevant.
[...]

PR #607, opened 2022-01-31
Add a progressive backoff mechanism for barriers.

devinamatthews:
The mechanism(s) used for backoff must be defined by the configuration or configuration family. Up to three successive mechanisms are supported (e.g. pause/sched_yield/sleep).

Example (in `config/<family>/bli_family_<family>.h`):
```C++
#define BLIS_BARRIER_BACKOFF_1 30    // number of times to loop
#define BLIS_BARRIER_YIELD_1 pause();pause();pause();

//#define BLIS_BARRIER_BACKOFF_2    not needed, the last mechanism is looped forever 
#define BLIS_BARRIER_YIELD_2 sched_yield();
```

This PR supplants #82 and potentially addresses #604 (family-specific behavior still needs to be decided upon and implemented).

Families needing implementations:
- [ ] Intel
- [ ] AMD (same as Intel?)
- [ ] ARM (v7 same as v8?)
- [ ] POWER

### DIFF
--- a/frame/thread/bli_thrcomm.c
+++ b/frame/thread/bli_thrcomm.c
@@ -41,7 +41,7 @@ void* bli_thrcomm_bcast
        void*      to_send,
        thrcomm_t* comm
      )
-{   
+{
 	if ( comm == NULL || comm->n_threads == 1 ) return to_send;
 
 	if ( id == 0 ) comm->sent_object = to_send;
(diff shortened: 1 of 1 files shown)


jeffhammond:
Jim Cownie wrote about this topic here: https://cpufun.substack.com/p/to-sched_yield-or-not-to-sched_yield.

PR #608, opened 2022-02-03
Arm64 & ArmSVE: Remove Generic Stride Assembly

xrq-phys:
I removed asm-level generic stride support from `armv8a` & `armsve` kernels to cope with #583 .

- [x] Test for NEON.
- [x] Test for SVE.
- [x] ~Use predicates.~ Use predicates only in `m`-dimension.

PR #609, opened 2022-02-06
Some preparatory re-working of kernels

### LINKED ISSUE #259: slow generic implementation
I was assuming that BLIS is generally better than reference BLAS, so substituting the latter with BLIS OS packages I'm working on would always be sensible.  However, I found BLIS is more than two times slower for medium-sized  dgemm on x86_64/RHEL7 for a "generic" build compared with the system reference blas package (which should be built with -O2 -mtune=generic, not -O3). I can't usefully test an architecture without a tuned implementation, but I don't see any reason to think that would be much different, though I haven't looked into the gcc optimization.

Is that expected, or might it be something worth investigating?

devinamatthews:
1) All kernels are now stored in a single array in the cntx_t. No differentiation is made between level-1[fvm], level-3 etc. in the cntx_t code except for a few l3 helper functions kept for compatibility.
2) Likewise, all (boolean) kernel preferences and block sizes are combined into single arrays. Notably, this subsumes all of the sup-specific structures in the cntx_t into the generic ones. The function `bli_stor3_ukr` is provided to map from sup storage types to generic ukr IDs.
3) Rework context initialization functions to use a sentinel instead of an explicit count. This should help to prevent errors where kernels are added or removed but the count isn't updated. Also, these functions no longer allocate memory.
4) Add a header for each sub-configuration (specifically, singleton configurations, which excludes families), which defines MR/NR for each datatype as macros, and is only included when compiling reference kernels. Definitions may be omitted to use the default values. This will allow reference kernels (including user kernels) to use these macros to generate optimized code. As an example, the reference gemm implementation achieves ~85-90% of the optimized performance on haswell/gcc-11 (MacBook Pro) when these values are available. Optimized kernels are assumed to have these values hard-coded. (NOTE: `armsve` is an oddball because MR is determined at runtime. Reference kernels will not be able to use a fixed value of MR/NR for optimization).

Fixes #259.


devinamatthews:
I should note that gcc-11 emits some warnings for `bli_gemm_ref` that I can't seem to get rid of, despite what the docs say about `#pragma GCC diagnostic ignore`.

devinamatthews:
@xrq-phys I'm debugging the ARM issues with this PR and all of the sudden I came across this error (clang12 on aarch64 VM):

```
kernels/armsve/3/bli_gemm_armsve_asm_d2vx10_unindexed.c:103:53: fatal error: invalid symbol redefinition
"                                                 \n\t"
                                                    ^
<inline asm>:20:3: note: instantiated into assembly here
         LOAD_ABC:                                       
         ^
1 error generated.
```

It seems to compile fine in Travis so I'm at a loss.

devinamatthews:
Clang10 works fine. I really hate clang sometimes (like, most of the time).

ref_kernels/3/bli_gemm_ref.c
```diff
@@ -103,14 +104,15 @@ void PASTEMAC3(ch,opname,arch,suf) \
...
+	if ( cs_c == 1 && m == mr && n == nr ) \
 	{ \
-		/* C is row-stored. */ \
+		/* C is row-stored and a full tile. */ \
 \
 		if ( PASTEMAC(ch,eq0)( *beta ) ) \
 		{ \
-			for ( dim_t i = 0; i < m; ++i ) \
-			for ( dim_t j = 0; j < n; ++j ) \
+			for ( dim_t i = 0; i < mr ++i ) \
```
hominhquan: missing ";" here

devinamatthews: Noted. I've got it fixed locally.

fgvanzee:
@devinamatthews Is this ready for me to review? (I see that Travis and AppVeyor are not yet satisfied.)

devinamatthews:
I think so. It has a good chance of passing Travis after eebe527 but I also merged #608 in and am testing in qemu now.

devinamatthews:
@fgvanzee All good.

devinamatthews:
@fgvanzee you did break it 😄 

I need to push another commit before you merge anyways, as I noticed that my rewrite of the packm_1er reference kernel has a bug.

devinamatthews:
@fgvanzee I will fix and repush

fgvanzee:
Ugh. I ran `make check`! How did *every* test fail? :confused:

PR #610, opened 2022-02-09
Simplify and rewrite reference packing kernels.

devinamatthews:
The array of reference packing kernels (0--31) are replaced by exactly two kernels for each config/datatype combination, one to pack MRxK micropanels and one to pack NRxK micropanels. **IMPORTANT**: the "bb" reference kernels have been merged into the "standard" kernels (packm [incl. 1er and unpackm], gemm, trsm, gemmtrsm). This replication factor is controlled by BLIS_BB[MN]_[sdcz] etc. Power9/10 need testing since only a replication factor of 1 has been tested. armsve also needs testing since the MR value isn't available as a macro.

~Depends on #609.~


fgvanzee:
So what was the source of the breakage in #609?

fgvanzee:
> So what was the source of the breakage in #609?

Nevermind, just saw your email.

fgvanzee:
BTW, we should update the `KernelsHowTo` documentation to reflect the changes to the `cntx_t` APIs.

devinamatthews:
That document will need many updates when I'm done with it.

devinamatthews:
@fgvanzee this should be ready to merge now, whenever 0.9.0 is out.

fgvanzee:
Cool. Thanks for the update.

PR #614, opened 2022-02-21
Add armsve to arm64 Metaconfig

### LINKED ISSUE #612: Integrate armsve in arm64 family : runtime detection and kernel selection
I believe the current arm64 config option will not include the armsve kernels.
```
configure: configuration 'arm64' is registered.
configure: 'arm64' is defined as having the following sub-configurations:
configure:    firestorm thunderx2 cortexa57 cortexa53 generic
configure:    armv8a generic
configure:   'firestorm' is registered...and exists.
configure:   'thunderx2' is registered...and exists.
configure:   'cortexa57' is registered...and exists.
configure:   'cortexa53' is registered...and exists.
configure:   'generic' is registered...and exists.
```

With new processors getting on the market, with SVE, it might be interesting to include the armsve optimization in the arm64 family, and to support the detection of sve at runtime.

@devinamatthews @jeffhammond

xrq-phys:
Can be blocked w/ the current `./configure` file if unsupported by the compiler.

Resolves #612

### DIFF
--- a/config/arm64/bli_family_arm64.h
+++ b/config/arm64/bli_family_arm64.h
@@ -39,7 +39,18 @@
 // -- MEMORY ALLOCATION --------------------------------------------------------
 
 #define BLIS_SIMD_ALIGN_SIZE 16
+#define BLIS_SIMD_MAX_NUM_REGISTERS 32
 
+// SVE-specific configs.
+#define N_L1_SVE_DEFAULT 64
+#define W_L1_SVE_DEFAULT 4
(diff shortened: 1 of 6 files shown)


devinamatthews:
@xrq-phys to be clear: configuring with `arm64` has been tested and works on both SVE and non-SVE archs?

xrq-phys:
@devinamatthews

Sorry made a mistake in last commit.

Now yes. I screened out SVE instructions on non-SVE machines.

Issue #613, opened 2022-02-17
armsve : non-multiple of NR for one or more datatypes


egaudry:
Running https://github.com/flame/blis/commit/c9700f369aa84fc00f36c4b817ffb7dab72b865d on an ARM sve-enabled processor, fails with
```
libblis: frame/base/bli_gks.c (line 451):
libblis: Default NC is non-multiple of NR for one or more datatypes.
libblis: Aborting.
```

devinamatthews:
Can you confirm that there is not a failure using the commit's parent [ee9ff98](https://github.com/flame/blis/commit/ee9ff988c49f16696679d4c6cd3dcfcac7295be7)?

egaudry:
I used
35195bb5cea5d99eb3eaf41e3815137d14ceb52d
9be97c150e19fa58bca30cb993a6509ae21e2025
and master (c9700f369aa84fc00f36c4b817ffb7dab72b865d) and got the same error with the 3 of them.

Could this be linked to some hard-coded SVE units size ? Sorry for the naive question.

devinamatthews:
@xrq-phys any idea about this?

xrq-phys:
Wield... [This line](https://github.com/flame/blis/blob/c9700f369aa84fc00f36c4b817ffb7dab72b865d/kernels/armsve/3/bli_armsve_utils.c#L81) should guarantee things' working.

xrq-phys:
I wonder whether you could try [this artifact](https://github.com/JuliaBinaryWrappers/blis_jll.jl/releases/tag/blis-v0.8.1%2B2) (the aarch64-linux one).

Perhaps it'd help eliminating the compiler's influence.

egaudry:
@xrq-phys I tried this however it seems some symbols are missing:
dsyr2k_ (https://github.com/JuliaBinaryWrappers/blis_jll.jl/releases/download/blis-v0.8.1%2B2/blis.v0.8.1.aarch64-linux-gnu.tar.gz).

If you let know how you build it, I can try on my own and activate what I need.

devinamatthews:
@fgvanzee is there a "verbose" mode that can be enabled at runtime to let us see what the actual values are? Failing that, @egaudry you could always add print statements to the file indicated by @xrq-phys (bli_armsve_utils.c).

xrq-phys:
Oops. I forgot there's a symbol mangling there.

On current master, in fact it should be OK to built via:
`configure armsve; make -j$NPROC`.

[...]

PR #615, opened 2022-02-22
ArmSVE Ensure Non-zero Block Size

xrq-phys:
Prompted by #613


devinamatthews:
@xrq-phys what kind of conditions lead to very small NC or MC?

xrq-phys:
Currently default L1/L2 size merely uses data from A64FX. Other VLs might cause somehow inconsistent block size.

It also acts as safeguard around user-specified L1/L2 environment variables.

PR #617, opened 2022-02-27
Revamp how user-specifiable tools and their environment variables (CC, CXX, FC, PYTHON, AR, RANLIB) are handled in configure

fgvanzee:
Make several changes to the way `configure` handles its search for `CC`, `CXX`, `FC`, `PYTHON`, `AR`, and `RANLIB`:
- Consolidates handling of all of the above tools into one `bash` function, `select_tool_w_env()`.
- If the user specifies a tool via an environment variable (e.g. `CC=gcc`) and that tool does not seem valid, print an error message and abort `configure`, unless the tool is optional (e.g. `CXX` or `FC`), in which case a warning message is printed instead.
- The definition of "seems valid" above amounts to:
   - responding to at least one of a basic set of command line options (e.g. `--version`, `-V`, `-h`) if the `os_name` is `Linux` (since GNU tools tend to respond to flags such as `--version`) **or** if the tool in question is `CC`, `CXX`, `FC`, or `PYTHON` (which tend to respond to the expected flags regardless of OS)
   - the binary merely existing for `AR`, and `RANLIB` on Darwin/OSX/BSD. (These OSes tend to have non-GNU versions of `ar` and `ranlib`, which often do not respond to `--version` and friends.)

### DIFF
--- a/configure
+++ b/configure
@@ -1079,44 +1079,6 @@ get_ranlib_search_list()
 	echo "${list}"
 }
 
-select_tool()
-{
-	local search_list CC_env the_cc cc
-
-	# This is the list of compilers/tools to search for, and the order in
-	# which to search for them.
-	search_list=$1
-
-	# The environment variable associated with the compiler/tool type we
-	# are searching (e.g. CC, CXX, PYTHON).
-	#CC_env=$2
-
-	# If CC_env contains something, add it to the beginning of our default
-	# search list.
-	#if [ -n "${CC_env}" ]; then
-	#	search_list="${CC_env} ${search_list}"
-	#fi
-
-	# Initialize our selected compiler/tool to empty.
-	the_cc=""
-
-	# Try each compiler/tool in the list and select the first one we find that
-	# works.
-	for cc in ${search_list}; do
-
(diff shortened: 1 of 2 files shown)


fgvanzee:
@devinamatthews Would you mind trying this out to make sure it behaves as you expected?

devinamatthews:
Yes, works great. I was able to fool it:

```
configure: user specified a C compiler via CC (ls).
configure: ls exists and appears to work.
configure: using 'ls' as C compiler.
```

but who cares. Checking for existence is plenty to guard against common issues (typos etc.).

fgvanzee:
> Yes, works great. I was able to fool it:
> 
> ```
> configure: user specified a C compiler via CC (ls).
> configure: ls exists and appears to work.
> configure: using 'ls' as C compiler.
> ```
> 
> but who cares. Checking for existence is plenty to guard against common issues (typos etc.).

Lol. Perhaps I should change that language to instead indicate that it "exists and responds to basic flags." Except that sometimes it actually just means it exists (and nothing else... looking at you, `ar` and `ranlib` on OSX).

devinamatthews:
Ship it.

PR #626, opened 2022-04-21
[WIP] [DONTMERGE] Optional offloading to AMD GPUs

iotamudelta:
This initial work via opt-in configure option enables offloading of some sgemm and dgemm operations to AMD GPUs via AMD's rocBLAS. It hence requires a working ROCm software stack and ROCm-enabled accelerator.

After enabling offloading capability, the default is "never offload". Offloading can be controlled through three environment variables:
BLIS_OFFLOAD=[never,always,thresh] - thresh enables threshold-dependent offloading
BLIS_OFFLOAD_SGEMM_THRESH=$number1 the threshold of M*N size of sgemm after which offloading should be attempted - must be specified
BLIS_OFFLOAD_DGEMM_THRESH=$number2 the threshold of M*N size of dgemm after which offloading should be attempted - must be specified

WIP:
* much more testing :-)
* improve formatting correctness - my astylerc isn't fully correct yet

Currently known limitations:
* offloading decision is made purely based on M*N size of gemm in conjunction w/ user-controlled thresholds (or always/never offload)
* rocBLAS is initialized w/ default settings - it'll hence use the first enumerated accelerator in a system and default stream

Future work:
* offloading of integer and complex gemms can be supported
* better offloading decision engine and performance model with less user input required

### DIFF
--- a/build/bli_config.h.in
+++ b/build/bli_config.h.in
@@ -183,5 +183,10 @@
 #define BLIS_DISABLE_COMPLEX_RETURN_INTEL
 #endif
 
+#if @enable_amd_offload@
+#define BLIS_ENABLE_AMD_OFFLOAD
+#else
+#define BLIS_DISABLE_AMD_OFFLOAD
+#endif
 
 #endif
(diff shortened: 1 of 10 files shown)


jeffhammond:
Why not make it a draft commit if you don't want it merged?

iotamudelta:
@jeffhammond should be ready for merge soon after WIP items one - and I'm happy to get any functional reviews already.

Issue #588, opened 2021-12-20
[aarch64] possible issue with atomic barrier and generic implementation (lack of good atomic support on generic kernel ?)


egaudry:
I've been observing a strange behavior when launching some job on an Ampere Altra processor (ARM neoverse-n1 based design) with multiple threads. The symptoms remain the same with both pthread and openmp threading : when launching the computation with multiple threads, the computation is likely to deadlock when a dgemm_ request is made.

By looking at the stracktrace, when the dgemm_ call is made, I can see an awful lot of threads there (see extract below, with the thread number). In this example, I was running with 64 threads in BLIS (with a 80-core processor).

```
Thread 1103 (Thread 0xffe2b1fcc5f0 (LWP 3186538) "actranpy"):
#0  0x0000ffffb3f9a018 in bli_thrcomm_barrier_atomic () from /bin/../lib/libblis.so.4
#1  0x0000ffffb3f99f8c in bli_thrcomm_bcast () from /bin/../lib/libblis.so.4
#2  0x0000ffffb3eee56c in bli_packm_alloc () from /bin/../lib/libblis.so.4
#3  0x0000ffffb3eefe2c in bli_packm_init () from /bin/../lib/libblis.so.4
#4  0x0000ffffb3eee638 in bli_packm_blk_var1 () from /bin/../lib/libblis.so.4
#5  0x0000ffffb3eefecc in bli_packm_int () from /bin/../lib/libblis.so.4
#6  0x0000ffffb3f1b784 in bli_l3_packb () from /bin/../lib/libblis.so.4
#7  0x0000ffffb3f1a398 in bli_l3_int () from /bin/../lib/libblis.so.4
#8  0x0000ffffb3f2d850 in bli_gemm_blk_var3 () from /bin/../lib/libblis.so.4
#9  0x0000ffffb3f1a398 in bli_l3_int () from /bin/../lib/libblis.so.4
#10 0x0000ffffb3f2d6fc in bli_gemm_blk_var2 () from /bin/../lib/libblis.so.4
#11 0x0000ffffb3f1a398 in bli_l3_int () from /bin/../lib/libblis.so.4
#12 0x0000ffffb3f99a30 in bli_l3_thread_entry () from /bin/../lib/libblis.so.4
#13 0x0000ffffb3627898 in start_thread () from /lib64/libpthread.so.0
#14 0x0000ffffb3571ddc in thread_start () from /lib64/libc.so.6
[...]
[...]

PR #630, opened 2022-05-03
add method to query HW size

jeffhammond:
This is my first step towards addressing the oversubscription encountered in https://github.com/flame/blis/issues/588.

This adds a utility to figure out how many CPUs (hardware threads) are available in the affinity mask of the calling context.  This is a stronger restriction than the number of available software threads.

@egaudry

Signed-off-by: Jeff Hammond <jehammond@nvidia.com>


devinamatthews:
Sorry @jeffhammond I hijacked the PR description to flesh out some ideas related to the larger problem of thread pinning.

jeffhammond:
To be clear, final version will not abort. But I could not figure out how to set BLIS threading variables correctly. That's why there's a preprocessing warning saying "please help me".

devinamatthews:
@jeffhammond once we're in the OMP region, it's kind of too late to do anything except, as in `bli_l3_thread_decorator_thread_check`, detect that we have only one thread when we expected more (e.g. if nested OMP is disabled). I think this check would have to come before the OMP region and initialization of the global communicator.

devinamatthews:
@jeffhammond can you walk me through the logic of the PR as implemented? I'm not sure how the CPU mask alone can reliably be used to detect oversubscription.

frame/thread/bli_l3_decor_openmp.c
```diff
@@ -199,6 +200,12 @@ void bli_l3_thread_decorator_thread_check
      )
 {
 	dim_t n_threads_real = omp_get_num_threads();
+        dim_t n_threads_hwmask;
+        if ( omp_in_parallel() ) {
```
jeffhammond, line 204: @devinamatthews here, we look for the number of HW threads available to the thread if in OpenMP already, or the process, if not in OpenMP already, and if the user is trying to use more SW threads than HW threads, we abort.

frame/thread/bli_l3_decor_openmp.c
```diff
@@ -241,6 +248,27 @@ void bli_l3_thread_decorator_thread_check
...
+	{
+                bli_print_msg( "The affinity mask on this process does not have "
+                               "enough HW threads for your requested SW threads.",
+                               __FILE__, __LINE__ );
+                bli_abort();
+
+		bli_thrcomm_init( n_threads_hwmask, gl_comm );
+		bli_rntm_set_num_threads_only( n_threads_hwmask, rntm );
+#warning HELP ME HERE
+		bli_rntm_set_ways_only( 1, 1, 1, 1, 1, rntm );
```
jeffhammond, line 268: @devinamatthews this is the part where i need help. i can just do serialization, but i was thinking about trying to do slightly better, e.g. if i have 40 cores and the user wants to run on 80 threads, i can set things up to use 40 threads properly.

frame/thread/bli_l3_decor_openmp.c
devinamatthews: Right, but this requires the user to have set the CPU mask (or OMP affinity?) correctly. What if I just run BLIS in OpenMP (with nesting) and set both OMP_NUM_THREADS and BLIS_NUM_THREADS equal to the number of cores? Or is this only meant to guard against a more specific use case?

jeffhammond:
the affinity mask HW thread count is compared to the user-specified SW thread count.

frame/thread/bli_l3_decor_openmp.c
devinamatthews: Reducing the number of active threads after thread creation is possible, but not easy. If this logic can go before the OpenMP region then things are much easier.

devinamatthews:
Ah OK, so the user could in principle reduce the number of HW cores "visible" to BLIS by setting a mask but if no action is taken then BLIS just sees all cores. OK.

devinamatthews:
Here are my suggested solutions:

1. Maintain a process-wide atomic `int` which counts the number of active threads. Abort (or warn etc.) if this exceeds the number of HW cores.
2. Prepare a set of tokens, one for each HW core. Threads check out these tokens when created. Abort if no tokens are available. This approach can be extended to address thread pinning since each token can indicate a particluar physical core.

devinamatthews:
Also, I suggest an environment variable `BLIS_OVERSUBSCRIBE`. The values `ABORT` and `WARN` would do what you expect and any other value ignores oversubscription. Maybe also a `SERIALIZE` option?

jeffhammond:
The token idea doesn't work if BLIS runs alongside other things that use threads. But I'll see what I can do. 

I disagree that nesting is the higher priority. The bug that motivated all of this was not because of nested OpenMP.

devinamatthews:
@egaudry @jeffhammond the issue in #588 was explained [here](https://github.com/flame/blis/issues/588#issuecomment-1021385664). How is that not nested OMP?

jeffhammond:
Okay it was nested open but also affinity. I can fix them both.

egaudry:
Sorry, I haven't had the time to check on my side.

@jeffhammond is correct that the issue was seen because an affinity mask was set
@devinamatthews is also correct that the issue originated from the oversubscription

to clarify, the slowdown occurred because more thread than available physical cores where used in BLIS within a region protected with an openmp lock.
this could happen as well with oversubscription in case of a nested omp loop, with more thread being used than physically available cores.

PR #632, opened 2022-05-11
Add support for `c_next` in the `auxinfo_t` struct.

fgvanzee:
This branch contains preliminary support for a new `.c_next` field within the `auxinfo_t` struct. It is fully implemented for `gemm`. Caveats:
- The "wrap-around" address computation for the edge cases is not yet verified (but should be close to correct).
- For now, only the `gemm` macrokernel (`bli_gemm_ker_var2()`) sets the `.c_next` field. The `gemmt`, `trmm`, and `trsm` macrokernels are (for now) oblivious.

(h/t to @devinamatthews and AMD for their contributions to this feature)


fgvanzee:
**Note to self:** Credit LeickR in the final squashed commit log.

PR #634, opened 2022-05-18
Define `bli_pthread_switch_t` + matching API

fgvanzee:
This PR implements a pthread-like switch type that can be used to safely move between two states. Similar to `pthread_once()`, the user-supplied init function will never be invoked in immediate succession, even if the caller attempts to turn the switch "on" repeatedly. However, unlike `pthread_once()`, a switch may be re-initialized (turned on) as long as it is first *de-initialized* (turned off).

@devinamatthews Let me know if you see anything problematic.

### DIFF
--- a/frame/thread/bli_pthread.c
+++ b/frame/thread/bli_pthread.c
@@ -693,3 +693,93 @@ int bli_pthread_barrier_wait
 
 #endif
 
+// -- Non-standard extensions --------------------------------------------------
+
+// -- pthread_switch --
+
+int bli_pthread_switch_on
+     (
+       bli_pthread_switch_t* sw,
+       int                 (*init)(void)
+     )
+{
+	// NOTE: This function assumes that init() will return 0 on success;
+	// otherwise, it will return some other integer. If the function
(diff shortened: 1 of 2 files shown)


fgvanzee:
Honestly, this would seem so useful for things like library initialization that I'm surprised it's not part of POSIX threads already.

devinamatthews:
👍

fgvanzee:
Whoops. Looks like I reversed the conditional for changing the switch:
```c
	// If the init() function succeeded, turn the switch on;
	// otherwise, leave the switch off.
	if ( r_val != 0 )
		sw->status = 1;
```
Should be `r_val == 0`, since `0` here means success.

hominhquan:
Just out of curiosity,  can this help removing the need of `pthread_once()` ?

(I would squash the two commits together)

fgvanzee:
> Just out of curiosity, can this help removing the need of `pthread_once()` ?
> 
> (I would squash the two commits together)

That is where we are headed, yes. :slightly_smiling_face:

hominhquan:
:+1:

PR #636, opened 2022-06-23
Use BLIS_ENABLE_COMPLEX_RETURN_INTEL in blastest files

bartoldeman:
cblat1 and zblat1 tests crash with --complex-return=intel;
in that case the original f2c'ed code is actually correct.

Adapt run-f2c.sh to enable conditional compilation; as
recursive-sed.sh is not included I approximated it using
```
#!/bin/bash
sed -i "$2" $(echo $4)
```
to test the changed sed commands in run-f2c.sh.

### DIFF
--- a/blastest/src/cblat1.c
+++ b/blastest/src/cblat1.c
@@ -475,11 +475,23 @@ static real c_b52 = 0.f;
     integer mx, my;
     complex cdot[1];
     integer lenx, leny;
-    extern /* Complex */ complex cdotc_(integer *, complex *, integer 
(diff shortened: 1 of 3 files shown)


fgvanzee:
> as recursive-sed.sh is not included

@bartoldeman The script is actually still available! I've since renamed it to `recu-sed.sh` for brevity. It's in the `build` directory, although I also keep a copy in my `~/bin`. :)

fgvanzee:
Thanks again for your patience, @bartoldeman.

PR #654, opened 2022-08-18
Changed all QuickStart scripts to use .sh suffix.

jdiamondGitHub:
Details:
- Our new QuickStart bash scripts had filenames initially ending in
  .bash.  We have changed them all to .sh, both internally and
  externally.
- Reverted the Altra subconfig QuickStart to using 1.07/1.08 as the
  demarcation where firmware changed affinity maps.
- Minor fixes to AltraMax tutorial where it had previously still
  used the Altra configuration.


jdiamondGitHub:
Field, you are our absolutely favorite developer on git that keeps an open source project that we send PRs to. :)

jdiamondGitHub:
This is my first attempt to send you a PR in a way that will be (reasonably) clean?  I created a new tiny branch called "small_update", and am PRing my changes from there.  You will also be thrilled to know that we Field-ified all of our bash scripts by changing the endings. (And Leick is supporting the ancient 1970s standard of using .sh, because .bash is only used by a small, super elite group of super Linux developers.)

PR #650, opened 2022-08-01
Arm64 Row-Major d8x6

xrq-phys:
Basically it's a templated re-implementation of current column-major d6x8's transpose.

Shall I test out `gemmtrsm`?

PR #655, opened 2022-08-19
Arm64 DGEMMSUP with Extended MR&NR

xrq-phys:
Added extended `MR/NR` support for `kernel/armv8a/3/sup`.

As num of registers in NEON is large in number and short in length, I'm here extending both `MR` and `NR`.
Way of approaching this is to represent C microtile in registers optionally in columns so for sizes like `6x7m`, `crr` is the kernel-level default with `rrr` supported through an in-register transpose.

- A few asm kernels are crafted for `rv` to complete this extended size support.
- For `rd` I'm still relying heavily in c-intrinsic kernels with branching so the performance might not be that nice. Sorry for that.
- As always, changes adopted into `firestorm` config.


fgvanzee:
Sorry for the delay in reviewing this, @xrq-phys. Looks like these changes are only active in the `firestorm` subconfig, so I see no harm in merging it. I assume you did some testing on your end?

xrq-phys:
Yes. I've done tests on my laptop.

Haven't tuned `sup`'s block size/threshold for other Arm64 subconfigs. Maybe another PR for Gravitons :D 

Btw this PR contains #650's change.

fgvanzee:
Thanks @xrq-phys!

And apologies for forgetting about #650. I've been so occupied with other BLIS-related projects. 

So once I merge this, will #650 be moot? Or do we still need to merge it?

xrq-phys:
No problem!

Yes #650 would be completely outdated once this is merged.

PR #658, opened 2022-09-07
Use kernel CFLAGS for 'kernels' subdirs in addons.

fgvanzee:
Details:
- Updated Makefile and common.mk so that the targeted configuration's kernel CFLAGS are applied to source files that are found in a 'kernels' subdirectory within an enabled addon. For now, this behavior only applies when the 'kernels' directory is at the top level of the addon directory structure. For example, if there is an addon named 'foobar', the source code must be located in addon/foobar/kernels/ in order for it to be compiled with the target configurations's kernel CFLAGS. Any other source code within addon/foobar/ will be compiled with general-purpose CFLAGS (the same ones that were used on all addon code prior to this commit). Thanks to AMD (esp. Mithun Mohan) for suggesting this change.
- Comment/whitespace updates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,8 +213,19 @@ MK_REFKERN_OBJS     := $(foreach arch, $(CONFIG_LIST), \
 MK_FRAME_OBJS       := $(call gen-obj-paths-from-src,$(FRAME_SRC_SUFS),$(MK_FRAME_SRC),$(FRAME_PATH),$(BASE_OBJ_FRAME_PATH))
 
 # Generate object file paths for the addon source code. If one or more addons
-# were not enabled a configure-time, this variable will we empty.
-MK_ADDON_OBJS       := $(call gen-obj-paths-from-src,$(ADDON_SRC_SUFS),$(MK_ADDON_SRC),$(ADDON_PATH),$(BASE_OBJ_ADDON_PATH))
+# were not enabled a configure-time, these variable will we empty.
+# NOTE: We separate the source and objects into kernel and non-kernel lists.
+MK_ADDON_KERS_SRC   := $(foreach addon, $(ADDON_LIST), \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -590,6 +602,23 @@ else
...
 endef
 
+# first argument: a configuration name from the union of config_list and
+# config_name, used to look up the CFLAGS to use during compilation.
+# second argument: the C99 addon file suffix being considered.
+# third argument: the name of the addon being considered.
+define make-c99-addon-kers-rule
+$(BASE_OBJ_ADDON_PATH)/$(3)/$(KERNELS_DIR)/%.o: $(ADDON_PATH)/$(3)/$(KERNELS_DIR)/%.$(2) $(BLIS_H_FLAT) $(ADDON_H99_FILES) $(MAKE_DEFS_MK_PATHS)
+ifeq ($(ENABLE_VERBOSE),yes)
+	$(CC) $(call get-addon-kernel-c99flags-for,$(1)) -c $$< -o $$@
```
MithunMohanKadavil, line 612: This rule is resulting in compilation errors since the header file paths are not part of call get-addon-kernel-c99flags-for. Adding get-addon-c99flags-for to this rule fixes the issue, but not sure if it is the right way to do it,

fgvanzee: You are right, Mithun! Thank you for identifying this issue and its fix. (It worked for me the first time only because I was using an dummy `.c` file in my addon `kernels` directory, and the file was empty, thus no need for the addon header paths.)

Please try my follow-up commit.

MithunMohanKadavil: I have tried the latest changes and it works perfectly. Thanks for the quick response.

fgvanzee: That's great news, Mithun. Thank you for your feedback!

PR #665, opened 2022-09-16
Implemented support for fat multithreading.

fgvanzee:
Details:
- Allow the user to configure BLIS in such a way that multiple threading implementations get compiled into the library, with one of those implementations chosen at runtime. For now, there are only three implementations available: OpenMP, pthreads, and single. (Here, 'single' merely refers to single-threaded mode.) The configure script now allows the user to give the `-t` option with a comma-separated list of values, such as `-t openmp,pthreads`. The first value in the list will always be the default at library initialization time, and `single` is always silently appended to the end of the list. The user can specify which implementation should execute in one of three ways: by setting the `BLIS_THREAD_IMPL` environment variable prior to launch; by calling the `bli_thread_set_thread_impl()` global runtime API; or by encoding their choice into a `rntm_t` that is passed into one of the expert interfaces. Any of these three choices overrides the initialization-time default (i.e., the first value listed to the `-t` `configure` option). Requesting an implementation that was not compiled into the library will result in an error message followed by `bli_abort()`.
- Relocated the `auto` logic for the `-t` option from the top-level `Makefile` to the `configure` script. (Currently, this logic is pretty dumb, choosing `openmp` for gcc and icc, and `pthreads` for clang.)
- Defined a new `timpl_t` `enum` in `bli_type_defs.h`, with three valid values: `BLIS_SINGLE`, `BLIS_OPENMP`, `BLIS_POSIX`.
- Reorganized the `thrcomm_t` struct into a single definition with two preprocessor blocks, one each for additional fields needed by OpenMP and pthreads.
- Added `timpl_t` argument to `bli_thrcomm_bcast()`, `bli_thrcomm_barrier()`, `bli_thrcomm_init()`, and `bli_thrcomm_cleanup()`, which these functions need since they are now wrappers that choose the implementation- specific function corresponding to the currently enabled threading implementation.
- Added `rntm_t*` to `bli_thread_broadcast()`, `bli_thread_barrier()` so that those functions can pass the `timpl_t` value into `bli_thrcomm_bcast()` and `bli_thrcomm_barrier()`, respectively.
- Defined `bli_env_get_str()` in `bli_env.c` to allow the querying of `BLIS_THREAD_IMPL` (which, unlike `BLIS_NUM_THREADS` and friends, is expected to be a string).
- Defined `bli_thread_get_thread_impl()`, `bli_thread_set_thread_impl()` to get and set the current threading implementation at runtime.
- Defined `bli_rntm_thread_impl()` and `bli_rntm_set_thread_impl()` to query and set the threading implementation within a `rntm_t`. Also choose `BLIS_SINGLE` as the default value when initializing `rntm_t` structs. 
- Added `bli_info_get_*()` functions to query whether OpenMP or pthreads would be chosen as the default at init-time. Note that this only tests whether OpenMP or pthreads is the first implementation in the list passed to the threading configure option (`-t`) and is *not* the same as querying which implementation is currently selected, since that can be influenced by `BLIS_THREAD_IMPL` and/or `bli_thread_set_thread_impl()`.
- Changed `l3int_t` to `l3int_ft`.
- Updated `docs/Multithreading.md` to document the new behavior.
- Updated `sandbox/gemmlike` and `addon/gemmd` to work with the new fat threading feature. This included a few bugfixes to bring the codes up to date, as necessary.
- Comment, whitespace updates.


fgvanzee:
Note: Cancelled most recent Travis CI due to trivial change.

devinamatthews:
Put `[ci skip]` on the commit message.
On Sep 21, 2022, at 6:29 PM, "Field G. Van Zee" ***@***.******@***.***>> wrote:

[EXTERNAL SENDER]

Note: Cancelled most recent Travis CI due to trivial change.

fgvanzee:
> Put `[ci skip]` on the commit message.

Could you elaborate? It's a little late to edit the commit message...

Also, I didn't skip the CI for the merge *back into* `master` (which passed), just on the branch prior to merge (which I tested locally prior to cancelling the CI).

PR #667, opened 2022-09-20
Parameterized test/3 drivers via command line args.

fgvanzee:
Details:
- Rewrote the drivers in `test/3`, the `Makefile`, and the `runme.sh` script so that most of the important parameters, including parameter combo, datatype, storage combo, induced method, problem size range, dimension bindings, number of repeats, and alpha/beta values can be passed in via command line arguments. (Previously, most of these parameters were hard-coded into the driver source, except a few that were hard-coded into the `Makefile`.) If no argument is given for any particular option, it will be assigned a sane default. Either way, the values employed at runtime will be printed to `stdout` before the performance data in a section that is commented out with `%` characters (which is used by matlab and octave for comments), unless the `-q` option is given, in which case the driver will proceed quietly and output only performance data. Thanks to Jeff Diamond for suggesting this CLI-based reimagining of the `test/3` drivers.
- Each driver provides extensive help via the `-h` option, with the help text tailored for the operation in question (e.g. `gemm`, `hemm`, `herk`, etc.). The driver also reminds the user which implementation it was linked to (e.g. blis, openblas, vendor, eigen).
- Defined `bli_env_get_str()` in `bli_env.c` so that the driver can query the environment variables that control multithreading.
- Minor changes to `bli_getopt()` in `bli_getopt.c` to accommodate the rewritten test drivers' command line parsing needs.


test/3/test_utils.h
```diff
@@ -0,0 +1,154 @@
...
+#endif
+
+	// Multithreading parameters: number of threads and affinity string.
+	char  nt_str[ MAX_STRING_SIZE + 1 ];
+	char  af_str[ MAX_STRING_SIZE + 1 ];
+
+	// Parameter combinations, datatype, operand storage combination,
+	// and induced method.
+	char  pc_str[ MAX_STRING_SIZE + 1 ];
+	char  dt_str[ MAX_STRING_SIZE + 1 ];
```
hominhquan, line 99: @fgvanzee Unless there will need to format these strings with `snprintf`, I think they can be changed to pointers:
```lang=c
char *dt_str;
```
and later use normal (pointer-)assignment, instead of `strncpy`:

```lang=c
params->dt_str =  GLOB_DEF_DT_STR;
```

That saves space and time. :)

fgvanzee: @hominhquan Hmm, let me make sure I understand. See line 90 of `test_utils.c`, for example:
```c
    case 'd':
    strncpy( params->dt_str, state.optarg, MAX_STRING_SIZE );
    gave_option_d = TRUE;
    break;
```
Here, we copy the argument string into the `params` struct if the `-d` option is given at runtime. Are you suggesting that we simply copy the pointer within `argv` rather than copy the contents, since `argv` is guaranteed to persist for the life of the program?

hominhquan: Oh I only saw `strncpy` from **static** strings (e.g. "gemm"/"hemm", `GLOB_DEF_DT_STR` and so on), and did not see the handle of getopt in `test_utils.c`. From what I see, the `getopt_t state;` is put in stack of the `parse_cl_params()` function, hence will be erased after the function return, so using any field of `state` is not safe. However, *iif* you are sure that their values always point to one of the `argv[]`, I think you can safely use `=` to "copy" them (i.e. pointer assignment).

Below is what is said in the C99 standard:
```
5.1.2.2.1 Program startup
...
The parameters argc and argv and the strings pointed to by the argv array shall
be modifiable by the program, and retain their last-stored values between program
startup and program termination
```

hominhquan: About argument processing, did you consider the GNU's [gengetopt](https://www.gnu.org/software/gengetopt/gengetopt.html) and the associated `.ggo` config file.

fgvanzee: No, I did not know about gengetopt, but thank you for the reference. I'll look into it for potential future usage.

So it sounds like you are in (cautious) agreement that I can cut out the string copy in favor of string pointer assignment. I'll test it out and see if anything breaks! :)

fgvanzee: @hominhquan Hmm, I may need to change the `GLOB_` string macros to global variables. I am concerned because each driver file defines its own `init_def_params()` function, which (after the change you proposed) would assign the address of the string to a field in the `params` struct. But the string is currently passed in as a string literal whose allocation would presumably disappear after `init_def_params()` returns. Changing the string macros to global variables would ensure that they persist. Thoughts?

hominhquan: > need to change the GLOB_ string macros to global variables

No, you won't need to. Just do something like:
```
char *msg = "Hello world";
```
In fact, any static string will be put in `.rodata` section - a kind of global "space" of the compiled ELF. So, the `msg` above will receive the address to the "global" string "Hello world". 

Below is a more handy example:
```
$ cat hello.c
// msg will receive address from .rodata section
const char *msg = "God save the king";

// idem, this function will return address of string in .rodata
const char* get_string_addr() {
    return "Hello world";
}

$ gcc -O2 -c hello.c

$ objdump -Dsr hello.o 

hello.o:     file format elf64-x86-64

Contents of section .text:
 0000 b8000000 00c3                        ......          
Contents of section .data:
 0000 00000000 00000000                    ........        
Contents of section .rodata.str1.1:
 0000 48656c6c 6f20776f 726c6400 476f6420  Hello world.God 
 0010 73617665 20746865 206b696e 6700      save the king.  
Contents of section .comment:
 0000 00474343 3a202847 4e552920 342e382e  .GCC: (GNU) 4.8.
 0010 35203230 31353036 32332028 52656420  5 20150623 (Red 
 0020 48617420 342e382e 352d3339 2900      Hat 4.8.5-39).  
Contents of section .eh_frame:
 0000 14000000 00000000 017a5200 01781001  .........zR..x..
 0010 1b0c0708 90010000 14000000 1c000000  ................
 0020 00000000 06000000 00000000 00000000  ................

Disassembly of section .text:

0000000000000000 <get_string_addr>:
   0:	b8 00 00 00 00       	mov    $0x0,%eax
			1: R_X86_64_32	.rodata.str1.1
   5:	c3                   	retq   

Disassembly of section .data:

0000000000000000 <msg>:
	...
			0: R_X86_64_64	.rodata.str1.1+0xc

```

So, if you mind about the "storage" of any "global" string, they are already saved somewhere in the compiled ELF (or libblis.so), and any assignment from those strings will take their pointer.

fgvanzee: Thanks for that explanation, @hominhquan. While waiting for your response, I also spoke to @devinamatthews and @LeickR, and they said that while it would work as you suggest most of the time, on most systems/compilers, it would be safest to declare global `const char*` strings versus rely on the string to be in the `.rodata`. (I forget their reasoning, though. Perhaps one of them could remind me?)

fgvanzee:
FYI, commit 63227ea makes a few changes unrelated to @hominhquan's review.

test/3/test_utils.h
fgvanzee: @hominhquan Take a look at eb8e68d. I think it achieves all of your goals, albeit with global `const char*` variables rather than via the `.rodata`. But in principle, both approaches should use the same amount of space I think?

hominhquan: That looks good to me. Thanks.

hominhquan: > both approaches should use the same amount of space I think?

Declaring global `const char*` will take 8 bytes more for each pointer on a 64-bit arch, but it's ok, they are not too many, and for the safety reason.

fgvanzee: Great. Thank you so much for your feedback here, @hominhquan. I love it when others help make our code better. :)

hominhquan: No worries, keeping BLIS small and clean is also helpful to us, so I'm happy to contribute :)

hominhquan: I remember someone said "Convenience always comes with a price..." https://godbolt.org/z/6f7ETsf8x
(I'm not allergic to C++, I love template and its ability to produce clean code, to be used with moderation :) )

devinamatthews: Yeeeaaaahh, there's `<string>` and there's `<iostream>`... I'm with you on the latter.

devinamatthews: I use fmtlib for I/O.

PR #668, opened 2022-09-23
Add consistent NaN/Inf handling in sumsqv.

devinamatthews:
- If there is a NaN (either real or imaginary), then return a sum of NaN and unit scale.
- Else, if there is an Inf (either real or imaginary), then return a sum of +Inf and unit scale.
- Otherwise as normal.

### DIFF
--- a/frame/util/bli_util_unb_var1.c
+++ b/frame/util/bli_util_unb_var1.c
@@ -1100,11 +1100,11 @@ void PASTEMAC(ch,varname) \
 			scale_r = one_r; \
 		} \
 \
-        if ( bli_isnan( sumsq_r ) ) \
-        { \
-            chi1 += incx; \
-            continue; \
-        } \
+		if ( bli_isnan( sumsq_r ) ) \
+		{ \
+			chi1 += incx; \
+			continue; \
+		} \
 \
 		if ( bli_isinf( abs_chi1_r ) ) \
 		{ \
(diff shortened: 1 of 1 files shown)


devinamatthews:
@fgvanzee after CI should be ready t merge. I tested on what should be a representative enough set of test cases to give the desired behavior.

fgvanzee:
Thanks for acting on this so quickly, @devinamatthews.

frame/util/bli_util_unb_var1.c
```diff
@@ -1085,10 +1085,39 @@ void PASTEMAC(ch,varname) \
...
+			sumsq_r = abs_chi1_i; \
+			scale_r = one_r; \
+			break; \
+		} \
+\
+		if ( bli_isinf( abs_chi1_r ) ) \
+		{ \
+			sumsq_r = abs_chi1_r; \
+			scale_r = one_r; \
+			break; \
```
EleniVlachop: Assuming that in a vector x we have x[0] = inf and x[1] = NaN, then we break out of the for loop and return inf in sumsq_r. From the description in the commit message, I expected to return NaN. Could you please clarify which one is the correct return value?

devinamatthews: This was fixed in 92b9af48871ee85e831939243c467b9ba5250fdb.

EleniVlachop: In that case, since we do not break in case of NaN, then for x[0] = inf , x[1] = NaN, x[2]= inf, won't we overwrite sumsq_r with inf again?

devinamatthews:
There is a `continue` in the most recent version ([[here](google.com)](https://github.com/flame/blis/blob/92b9af48871ee85e831939243c467b9ba5250fdb/frame/util/bli_util_unb_var1.c#L1106)). Once a NaN is encountered this will trigger every iteration. That particular line could be a `break` but it doesn't make a difference to the result.

frame/util/bli_util_unb_var1.c
devinamatthews: There is a `continue` in the most recent version ([here](https://github.com/flame/blis/blob/92b9af48871ee85e831939243c467b9ba5250fdb/frame/util/bli_util_unb_var1.c#L1106)). Once a NaN is encountered this will trigger every iteration. That particular line could be a `break` but it doesn't make a difference to the result.

PR #669, opened 2022-09-29
Add an --enable-asan option.

devinamatthews:
This option (disbaled by default) enables compiling and linking with the Address Sanitizer library (ASan), via the -fsanitize=address flag supported by clang, gcc, and probably others. This flag is included for all files *except* optimized kernels, since it usually reguires an extra register which violates the constraints for many gemm microkernels.


devinamatthews:
@fgvanzee should be a simple PR

fgvanzee:
@devinamatthews In principle, do we want the address sanitizer flags included in the list of compiler flags that applications import, or only the list of flags used when BLIS is being compiled?

devinamatthews:
I didn't know the former existed, but yes it should probably go there.

fgvanzee:
> I didn't know the former existed, but yes it should probably go there.

From `common.mk`, line 205-ish:
```bash
# Define a separate function that will return appropriate flags for use by
# applications that want to use the same basic flags as those used when BLIS
# was compiled. (NOTE: This is the same as the $(get-frame-cflags-for ...)
# function, except that it omits two variables that contain flags exclusively
# for use when BLIS is being compiled/built: BUILD_CPPFLAGS, which contains a
# cpp macro that confirms that BLIS is being built; and BUILD_SYMFLAGS, which
# contains symbol export flags that are only needed when a shared library is
# being compiled/linked.)
get-user-cflags-for      = $(strip $(call load-var-for,COPTFLAGS,$(1)) \
                                   $(call get-noopt-cflags-for,$(1)) \
                            )
```
I asked because sometimes there are flags we don't want/need to expose to the user.

So, just to confirm: if BLIS is intentionally build with `-fsanitize=address`, you're saying that a program that links to that copy of BLIS would also want to use that flag?

devinamatthews:
It would need to in order to link properly.

fgvanzee:
> It would need to in order to link properly.

That would be taken care of via `LDFLAGS`, to which `-fsanitize=address` is already added. So the application wouldn't necessarily want it *during compilation*, yeah?

devinamatthews:
Yes, it most likely would.

devinamatthews:
But also, if a user is aware enough to turn this option on then they are probably also aware enough to add it to their application flags themselves.

fgvanzee:
> But also, if a user is aware enough to turn this option on then they are probably also aware enough to add it to their application flags themselves.

I can't tell if this is offering support for including the flag in the flags returned by `get-user-cflags-for()` (in addition to `LDFLAGS`, which we both agree on), or if you're saying it's not necessary after all since they would want to manually control whether it's used during their application's compilation.

devinamatthews:
I'm saying it probably doesn't make a lot of difference.

fgvanzee:
Thanks for clarifying. I'll leave it out for now, then (and update the comment).

fgvanzee:
Note: CI was indirectly cancelled by merging prior to finish. No issues were detected in subsequent CI tests.

PR #670, opened 2022-09-29
Fix some bugs in bli_pool.c

devinamatthews:
1. Add a check for pool exhaustion when freeing blocks. This detects double-free and other bad conditions without segfault.
2. Make sure to copy *all* block pointers when growing the pool size. Previously, checked-out block pointers were not copied, leading to the presence of uninitialized data.


devinamatthews:
@fgvanzee should be a simple PR.

PR #671, opened 2022-10-03
Reinstate sanity check in bli_pool_finalize.

devinamatthews:
Reinstate check for checked-out blocks upon finalization. A flag has been added to indicate that the pool is actually under reinitialization (where checked-out blocks are OK), which temporarily disables the check. A memory leak where blocks are not checked back in is now correctly detected upon exit.

### DIFF
--- a/frame/base/bli_apool.c
+++ b/frame/base/bli_apool.c
@@ -188,7 +188,7 @@ void bli_apool_free_block
 		if ( pool != NULL )
 		{
 			// Finalize the pool.
-			bli_pool_finalize( pool );
+			bli_pool_finalize( pool, FALSE );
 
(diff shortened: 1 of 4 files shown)


devinamatthews:
@fgvanzee another fix for bli_pool.

fgvanzee:
Can you talk a little bit more about what happened here? Was it that both the original check and your previous fix were incomplete, and we needed to somehow handle both?

devinamatthews:
The previous PR prevented a segfault when certain conditions happened (in my case, a memory leak). This PR reinstates the check necessary to detect the memory leak instead of exiting successfully. It also avoids a spurious check when reinitializing the pool in mid-computation.

fgvanzee:
Okay, I think I understand now. Thanks.

PR #680, opened 2022-10-31
Fix bug where multiple threads pack the same parts of each panel.

devinamatthews:
This was caused by different parts of the code using num_threads/thread_id vs. n_way/work_id. The fix was to standardize on the latter and provide a "fake" thrinfo_t sub-prenode in the thread control tree which consists of single-threaded teams. The single-team with multiple threads node is also required since it and only it can be used to do barriers and broadcasting (e.g. of the packed buffer pointer).

### DIFF
--- a/frame/3/bli_l3_thrinfo.c
+++ b/frame/3/bli_l3_thrinfo.c
@@ -77,17 +77,18 @@ void bli_l3_thrinfo_grow
 	thrinfo_t* thread_cur = bli_thrinfo_split( n_way, thread_par );
 	bli_thrinfo_set_sub_node( thread_cur, thread_par );
 
-    if ( bszid == BLIS_NO_PART )
-    {
-        // A hack: the packing code needs a thread communicator which represents
-        // a group of single-threaded teams working together. However, the
-        // "normal" packm thrinfo_t node has a single team with multiple threads.
-        // So, we create a sub-prenode on the thrinfo_t tree which splits this
-        // single team into multiple single-threaded teams.
-        const dim_t n_threads = bli_thrinfo_num_threads( thread_par );
+	if ( bszid == BLIS_NO_PART )
+	{
+		// A hack: the packing code needs a thread communicator which represents
+		// a group of single-member thread teams working cooperatively However,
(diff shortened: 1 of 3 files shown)


frame/3/bli_l3_thrinfo.c
```diff
@@ -77,7 +77,18 @@ void bli_l3_thrinfo_grow
 	thrinfo_t* thread_cur = bli_thrinfo_split( n_way, thread_par );
 	bli_thrinfo_set_sub_node( thread_cur, thread_par );
 
-	if ( sub_prenode != NULL )
+    if ( bszid == BLIS_NO_PART )
```
hominhquan, line 80: I wonder if there is a replacement to the condition `if ( bszid == BLIS_NO_PART )` (there are tons of it in BLIS) to something clearer, like:

```lang=c
const bool doing_packing = ...; // perhaps giving a new name to BLIS_NO_PART to mean "packing" ?
if ( doing_packing )
{
    ...
}
```

PS: you seem to have indent issues in some places, or it was the Github viewer ?

devinamatthews: We have lots of indent issues because I write spaces and then Field changes them to tabs. It's a process.

devinamatthews: But this is really a hack for now to fix performance and potential correctness issues. This will be replaced by something more robust eventually.

devinamatthews: Although do note that the only control tree nodes with BLIS_NO_PART at the moment are packing nodes.

hominhquan: > Although do note that the only control tree nodes with BLIS_NO_PART at the moment are packing nodes.

Yes, that's why I suggested to rename it to something more understandable. I recall myself the first time reading the code to find out what does `BLIS_NO_PART` mean.

fgvanzee: I was trying to future-proof the `enum` when naming it `BLIS_NO_PART` because I envisioned maybe someday there would be an operation that doesn't partition but doesn't pack either. I agree it's not super clear and I'm 100% open to it being improved.

PR #682, opened 2022-11-02
blis support for hpx

### LINKED ISSUE #681: provide support for the hpx runtime system
[HPX](https://github.com/STEllAR-GROUP/hpx) is an asynchronous many task runtime system used in high performance computing applications. The runtime system implements the ISO C++ parallelism specification and provides a user-space thread implementation. This feature request is a placeholder for a PR that provides BLIS a thread backend implementation using HPX. The HPX backend for BLIS reduces thread contention issues, caused by thread schedulers competing with each other, when implementing HPX applications that utilize BLIS.

ct-clmsn:
[HPX](https://github.com/STEllAR-GROUP/hpx) is an asynchronous many task runtime system used in high performance computing applications. The runtime implements the ISO C++ parallelism specification and provides a user-space thread implementation.

This PR provides BLIS a thread backend implementation using HPX and resolves feature request #681. The configuration script and makefiles have been updated to support an HPX build option. The examples directory has been expanded with modified versions of the existing suite of BLIS examples that are implemented using HPX. The addition of HPX support provides other developers an exemplar for integrating other C++ backends into BLIS.

Fixes #681.


ct-clmsn:
The following configurations have been tested with my development environment successfully:

- ./configure --enable-threading=none x86_64
- ./configure --enable-threading=single x86_64
- ./configure --enable-threading=openmp x86_64
- ./configure --enable-threading=pthreads x86_64
- ./configure --enable-threading=hpx x86_64
 
The modifications applied to `configure` and `Makefile` only impact the thread support portions of the build system. Unsure what the discrepancy is with my development environment and the CI solution. I could use assistance with the PR. Thank you in advance for your time and attention.

ct-clmsn:
The last commit fixes an issue in the hpx implementation of `examples/hpx/tapi/03level2.cpp` and removes automatic debug build support for examples in the same path.

devinamatthews:
I can review this shortly. Quick question @ct-clmsn does HPX require a C++ compiler for the bits you added in BLIS?

ct-clmsn:
@devinamatthews yep. The c++ compiler gets applied in the one part of the build where hpx is integrated.

devinamatthews:
@ct-clmsn why did you add an `hpx` directory in `examples`? It doesn't really seem necessary.

ct-clmsn:
@devinamatthews the hpx runtime has multiple techniques to get initialized. The technique applied in 'examples/hpx' is the most portable. The examples are in a sub-directory to avoid changing the existing versions. If you'd like to have the hpx versions merged in with the original suite with compile guards, I can get that done over the weekend.

devinamatthews:
I guess compiler guards would be OK. I don't think it's too important though since it isn't really necessary to have the examples compile with all threading models. They're just examples and mostly to look at not run.

ct-clmsn:
Good deal, I'll get those change made.

devinamatthews:
@ct-clmsn the testsuite binaries also need to be updated to initialize HPX. Can one use `hpx::init(nullptr, argc, argv)` to get a more `MPI_Initalize`-like experience without the `hpx_main` function?

ct-clmsn:
@devinamatthews there isa technique that can give you something similar to MPI_Init. A sample program of the technique is shown in the link below.

https://hpx-docs.stellar-group.org/latest/html/manual/starting_the_hpx_runtime.html#suspend-resume

devinamatthews:
OK, that should be much easier to wrap in compiler guards.

devinamatthews:
@ct-clmsn I created extern "C" functions for initialization and finalization of HPX:

```C

int main( int argc, char** argv )
{
#ifdef BLIS_ENABLE_HPX
    bli_thread_initialize_hpx( argc, argv );
#endif

	....

#ifdef BLIS_ENABLE_HPX
    return bli_thread_finalize_hpx();
#else
	return 0;
#endif
}
```

devinamatthews:
@ct-clmsn I pushed some new commits to finish this out. The testsuite passes for me on macOS (not easy to get HPX working there!). If the CI is still failing I'll check it out.

devinamatthews:
Ah, building *without* HPX is broken. Should be an easy fix.

ct-clmsn:
The initialization functions look great, thanks for putting those together!

ct-clmsn:
@devinamatthews did you need me to push on anything this week?

devinamatthews:
@ct-clmsn something is broken with `configure -t omp auto` if you want to fix that (I think it's something in the Makefile?).

ct-clmsn:
@devinamatthews sure thing. I'll check into it.

ct-clmsn:
@devinamatthews sorted the issue and did an update; *.cpp suffix files needed a build rule to get hpx integrated. that build rule was conditionally turned off when hpx was not enabled. issue should be fixed now.

devinamatthews:
@ct-clmsn we prefer not to require a C++ compiler to build the framework, so I tweaked your fix to just not build cpp files at all if HPX threading is not enabled. If it passes CI testing we can merge. I haven't added an HPX test to Travis because it looks like we'd need to build HPX from source... if you want to take a whack at putting such a test together please go ahead and open another PR.

devinamatthews (approved):

ct-clmsn:
@devinamatthews perfect! Thanks for the assistance!

PR #690, opened 2022-11-11
Fixed subtle barrier_fpa bug in bli_thrcomm.c.

fgvanzee:
Details:
- In `bli_thrcomm.c`, correctly initialize the `BLIS_OPENMP` element of the barrier function pointer array (`barrier_fpa`) to `NULL` when `BLIS_ENABLE_OPENMP` is *not* defined. Similarly, initialize the `BLIS_POSIX` element of `barrier_fpa` to `NULL` when `BLIS_ENABLE_PTHREADS` is not defined. This bug was introduced in a1a5a9b and was likely the result of an incomplete edit. The effects of the bug would have likely manifested when querying a `thrcomm_t` that was initialized with a `timpl_t` value corresponding to a threading implementation that was omitted from the `-t` option at configure-time.


devinamatthews:
Thx I think I also fixed this similarly in the HPX PR.

PR #693, opened 2022-11-20
Add RISC-V target

angsch:
Add the infrastructure for RISC-V and an optimized SGEMM kernel assuming VLEN=128.

The config is intentionally set up to be as generic as possible. The code can be run with the GNU RISC-V toolchain (compiler + qemu) available as Ubuntu packages.

The vector extension for application cores supports various vector lengths starting from 128 bits. The assembly kernel assumes VLEN=128, which is, for example, used in P670 or P470. If the vector length is not 128, a fallback is needed. For the time being, this is the generic implementation. There will likely be a few revisions and, at some point, a vector-length agnostic implementation should be added.

Is this case distinction in the C file acceptable or do you prefer specialized targets?


config/rv32iv/make_defs.mk
```diff
@@ -0,0 +1,90 @@
...
+CWARNFLAGS     := -Wall -Wno-unused-function -Wfatal-errors
+
+ifneq ($(DEBUG_TYPE),off)
+CDBGFLAGS      := -g
+endif
+
+ifeq ($(DEBUG_TYPE),noopt)
+COPTFLAGS      := -O0
+else
+COPTFLAGS      := -O2 -ftree-vectorize -march=rv32iav
```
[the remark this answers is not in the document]
angsch: - atomic instructions (`a` flag) are not longer required to be available in hardware since it is possible to link against a library that emulates the required instructions in software.
- `mabi` has been revised such that `rv[32,64]i` use the minimal mabi. `rv[32,64]iv` can pass floating-point values in registers, which corresponds to the default build of the toolchain and is compatible with other libraries.
- `march` is now set twice for `COPTFLAGS`. The rationale is to enable user-defined hardware extensions (e.g., `a`tomic, `c`ompressed instructions) that shall only apply to one type of build.

config/rv32i/make_defs.mk
```diff
@@ -0,0 +1,90 @@
...
+#CONFIGS_INCL   += $(THIS_CONFIG)
+
+#
+# --- Determine the C compiler and related flags ---
+#
+
+# NOTE: The build system will append these variables with various
+# general-purpose/configuration-agnostic flags in common.mk. You
+# may specify additional flags here as needed.
+CPPROCFLAGS    := -mabi=ilp32
```
leekillough, line 48: Should `-mabi` be added to `CMISCFLAGS` rather than preprocessor flags?

config/rv32i/make_defs.mk
```diff
@@ -0,0 +1,90 @@
...
+CWARNFLAGS     := -Wall -Wno-unused-function -Wfatal-errors
+
+ifneq ($(DEBUG_TYPE),off)
+CDBGFLAGS      := -g
+endif
+
+ifeq ($(DEBUG_TYPE),noopt)
+COPTFLAGS      := -O0 -march=rv32i
+else
+COPTFLAGS      := -O2 -march=rv32i
```
leekillough, line 60: `-march`  should probably be added to `CMISCFLAGS` as well, since it more global than optimization flags.

config/rv32iv/make_defs.mk
```diff
@@ -0,0 +1,92 @@
...
+#CONFIGS_INCL   += $(THIS_CONFIG)
+
+#
+# --- Determine the C compiler and related flags ---
+#
+
+# NOTE: The build system will append these variables with various
+# general-purpose/configuration-agnostic flags in common.mk. You
+# may specify additional flags here as needed.
+CPPROCFLAGS    := -mabi=ilp32d
```
leekillough, line 48: See above.

config/rv32iv/make_defs.mk
```diff
@@ -0,0 +1,92 @@
...
+CMISCFLAGS     :=
+CPICFLAGS      :=
+CWARNFLAGS     := -Wall -Wno-unused-function -Wfatal-errors
+
+ifneq ($(DEBUG_TYPE),off)
+CDBGFLAGS      := -g
+endif
+
+ifeq ($(DEBUG_TYPE),noopt)
+COPTFLAGS      := -O0 -march=rv32iv
```
leekillough, line 58: See above.

config/rv64gv/make_defs.mk
```diff
@@ -0,0 +1,90 @@
...
+#CONFIGS_INCL   += $(THIS_CONFIG)
+
+#
+# --- Determine the C compiler and related flags ---
+#
+
+# NOTE: The build system will append these variables with various
+# general-purpose/configuration-agnostic flags in common.mk. You
+# may specify additional flags here as needed.
+CPPROCFLAGS    := -mabi=lp64d
```
leekillough, line 48: See above.

config/rv64gv/make_defs.mk
```diff
@@ -0,0 +1,90 @@
...
+CWARNFLAGS     := -Wall -Wno-unused-function -Wfatal-errors
+
+ifneq ($(DEBUG_TYPE),off)
+CDBGFLAGS      := -g
+endif
+
+ifeq ($(DEBUG_TYPE),noopt)
+COPTFLAGS      := -O0
+else
+COPTFLAGS      := -O2 -ftree-vectorize -march=rv64gv
```
leekillough, line 60: See above.

config/rv64i/make_defs.mk
```diff
@@ -0,0 +1,90 @@
...
+#CONFIGS_INCL   += $(THIS_CONFIG)
+
+#
+# --- Determine the C compiler and related flags ---
+#
+
+# NOTE: The build system will append these variables with various
+# general-purpose/configuration-agnostic flags in common.mk. You
+# may specify additional flags here as needed.
+CPPROCFLAGS    := -mabi=lp64
```
leekillough, line 48: See above.

config/rv64i/make_defs.mk
```diff
@@ -0,0 +1,90 @@
...
+CMISCFLAGS     :=
+CPICFLAGS      :=
+CWARNFLAGS     := -Wall -Wno-unused-function -Wfatal-errors
+
+ifneq ($(DEBUG_TYPE),off)
+CDBGFLAGS      := -g
+endif
+
+ifeq ($(DEBUG_TYPE),noopt)
+COPTFLAGS      := -O0 -march=rv64i
```
leekillough, line 58: See above.

config/rv64iv/make_defs.mk
```diff
@@ -0,0 +1,94 @@
...
+#CONFIGS_INCL   += $(THIS_CONFIG)
+
+#
+# --- Determine the C compiler and related flags ---
+#
+
+# NOTE: The build system will append these variables with various
+# general-purpose/configuration-agnostic flags in common.mk. You
+# may specify additional flags here as needed.
+CPPROCFLAGS    := -mabi=lp64d -D_RV64
```
leekillough, line 48: See above.

config/rv64iv/make_defs.mk
```diff
@@ -0,0 +1,94 @@
...
+CMISCFLAGS     :=
+CPICFLAGS      :=
+CWARNFLAGS     := -Wall -Wno-unused-function -Wfatal-errors
+
+ifneq ($(DEBUG_TYPE),off)
+CDBGFLAGS      := -g
+endif
+
+ifeq ($(DEBUG_TYPE),noopt)
+COPTFLAGS      := -O0 -march=rv64iv
```
leekillough, line 58: See above.

config/rv64iv/make_defs.mk
```diff
@@ -0,0 +1,94 @@
...
+else
+# The latest build hits an 'internal compiler error'
+# when compiling the reference gemm kernels. Workout
+# via -march=rv64imv
+COPTFLAGS      := -O2 -ftree-vectorize -march=rv64iv
```
leekillough, line 63: See above.

config_registry
```diff
@@ -46,5 +46,13 @@ power10:     power10
 power9:      power9
 bgq:         bgq
 
+# RISC-V architectures.
+rv32i:       rv32i/rvi
+rv64i:       rv64i/rvi
+rv32iv:      rv32iv/rviv
+rv64iv:      rv64iv/rviv
+
+rv64gv:      rv64gv
```
leekillough, line 55: Is `rv64gv` supposed to remain as a special case?

frame/base/bli_arch.c
```diff
@@ -233,6 +233,24 @@ void bli_arch_set_id( void )
...
+		#endif
+		#ifdef BLIS_FAMILY_RV32IV
+		id = BLIS_ARCH_RV32IV;
+		#endif
+		#ifdef BLIS_FAMILY_RV64IV
+		id = BLIS_ARCH_RV64IV;
+		#endif
+
+		#ifdef BLIS_FAMILY_RV64GV
+		id = BLIS_ARCH_RV64GV;
```
leekillough, line 251: See above.

frame/base/bli_arch.c
```diff
@@ -284,6 +302,12 @@ static const char* config_name[ BLIS_NUM_ARCHS ] =
     "power7",
     "bgq",
 
+    "rv32i",
+    "rv64i",
+    "rv32iv",
+    "rv64iv",
+    "rv64gv",
```
leekillough, line 309: See above.

frame/base/bli_gks.c
```diff
@@ -208,6 +208,38 @@ void bli_gks_init( void )
...
+		bli_gks_register_cntx( BLIS_ARCH_RV64IV,      bli_cntx_init_rv64iv,
+		                                              bli_cntx_init_rv64iv_ref,
+		                                              bli_cntx_init_rv64iv_ind );
+#endif
+
+#ifdef BLIS_CONFIG_RV64GV
+		bli_gks_register_cntx( BLIS_ARCH_RV64GV,      bli_cntx_init_rv64gv,
+		                                              bli_cntx_init_rv64gv_ref,
+		                                              bli_cntx_init_rv64gv_ind );
+#endif
```
leekillough, line 241: See above.

frame/include/bli_arch_config.h
```diff
@@ -131,6 +131,25 @@ CNTX_INIT_PROTS( power7 )
...
+CNTX_INIT_PROTS( rv64i )
+#endif
+#ifdef BLIS_CONFIG_RV32IV
+CNTX_INIT_PROTS( rv32iv )
+#endif
+#ifdef BLIS_CONFIG_RV64IV
+CNTX_INIT_PROTS( rv64iv )
+#endif
+#ifdef BLIS_CONFIG_RV64GV
+CNTX_INIT_PROTS( rv64gv )
```
leekillough, line 149: See above.

kernels/rv64gv/3/bli_gemm_rv64gv_asm_s16x4.c
```diff
@@ -0,0 +1,123 @@
...
+                    ab[i + j * 16] += ai * bj;
+                }
+            }
+            a += 16;
+            b += 4;
+        }
+
+
+        for (dim_t i = 0; i < 16 * 4; i++) {
+            ab[i] = ab[i] * (*alpha);
```
leekillough, line 112: If `alpha == 0`, `ab[i]` should be zeroed regardless of `a[i]` and `b[i]`. GEMM conventions say that `alpha == 0` or `beta == 0` should short-circuit multiplications to `0` even in the presence of `NaN` entries.

kernels/rv64gv/3/bli_gemm_rv64gv_asm_s16x4.c
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        for (dim_t j = 0; j < n; j++) {
+            for (dim_t i = 0; i < m; i++) {
+                c[i * rs_c + j * cs_c] = c[i * rs_c + j * cs_c] * (*beta) + ab[i + j * 16];
```
leekillough, line 117: See comment above.

kernels/rv64gv/3/bli_sgemm_rv64gv_asm_16x4.S
```diff
@@ -0,0 +1,419 @@
+#define REALNAME bli_sgemm_rv64gv_ker_16x4
```
leekillough, line 1: `m4` macros are more powerful than C preprocessor macros.

Just a style thing at this point.

kernels/rv64gv/3/bli_sgemm_rv64gv_asm_16x4.S
```diff
@@ -0,0 +1,419 @@
...
+    vfmacc.vf AB31, B01, A30
+    vfmacc.vf AB32, B02, A30
+    vfmacc.vf AB33, B03, A30
+
+
+MULTIPLYALPHA:
+    flw ALPHA,(a1)
+    flw BETA, (a4)
+
+    // Multiply with alpha
```
leekillough, line 277: Need to short-circuit `alpha == 0` if not handled elsewhere.

kernels/rv64gv/3/bli_gemm_rv64gv_asm_s16x4.c
devinamatthews: `alpha == 0` is handled at a much higher level. The micro-kernels are only responsible for handling `beta == 0` and `k == 0`.

kernels/rviv/3/bli_dgemm_rviv_asm_4vx4.S
```diff
@@ -0,0 +1,417 @@
...
+    vfmacc.vf AB31, B01, A30
+    vfmacc.vf AB32, B02, A30
+    vfmacc.vf AB33, B03, A30
+
+
+MULTIPLYALPHA:
+    fld ALPHA,(a1)
+    fld BETA, (a4)
+
+    // Multiply with alpha
```
leekillough, line 276: Needs `alpha == 0` short-circuit.

kernels/rviv/3/bli_sgemm_rviv_asm_4vx4.S
```diff
@@ -0,0 +1,418 @@
...
+    vfmacc.vf AB31, B01, A30
+    vfmacc.vf AB32, B02, A30
+    vfmacc.vf AB33, B03, A30
+
+
+MULTIPLYALPHA:
+    flw ALPHA,(a1)
+    flw BETA, (a4)
+
+    // Multiply with alpha
```
leekillough, line 277: Needs `alpha == 0` short-circuit.

kernels/rviv/3/bli_zgemm_rviv_asm_4vx4.S
```diff
@@ -0,0 +1,779 @@
...
+    vfmsub.vf AB12_re, ALPHA_re, tmp2_re
+    vfmsub.vf AB13_re, ALPHA_re, tmp3_re
+    vfmadd.vf AB10_im, ALPHA_re, tmp0_im
+    vfmadd.vf AB11_im, ALPHA_re, tmp1_im
+    vfmadd.vf AB12_im, ALPHA_re, tmp2_im
+    vfmadd.vf AB13_im, ALPHA_re, tmp3_im
+
+    j MULTIPLYBETA
+
+ALPHAREAL:
```
leekillough, line 619: Needs `alpha == 0` short-circuit.

devinamatthews: @leekillough see my comment above re `alpha`.

leekillough:
@devinamatthews @angsch : Please see https://github.com/angsch/blis/pull/1 (hide whitespace to avoid noise)

angsch:
Summary of changes:

- Remove legacy target rv64gv with VLEN=128
- Revise Makefiles: Type (32 bit/64 bit) is propagated via define, add comment on atomic instructions, move `abi` and `march` to `CMISCFLAGS`
- Squash commits that do not have a value on their own.

Pending:
- Shortcut `k == 0`
- Activate assembly kernels for rv32iv target

angsch:
Summary of changes:

- Back to a working version that is in sync with master
- Remove hack to get the vector length
- Complete set of GEMM kernels compatible with RVV 1.0

Pending:
- CI
- flags for clang

.travis.yml
```diff
@@ -79,13 +93,27 @@ script:
...
+      export TESTSUITE_WRAPPER="$DIST_PATH/../toolchain/qemu-riscv32 -cpu rv32,vext_spec=v1.0,v=true,vlen=128 -B 0x100000";
+    fi
+- $DIST_PATH/configure -p `pwd`/../install -t $THR $BLD CC=$CC $CONF
 - pwd
 - ls -l
 - $CC --version
 - make -j 2
 - make install
-- $DIST_PATH/travis/cxx/cxx-test.sh $DIST_PATH $(ls -1 include)
+- if [ "BLD" = "" ]; then $DIST_PATH/travis/cxx/cxx-test.sh $DIST_PATH $(ls -1 include); fi
```
leekillough, line 116: `"$BLD"`

travis/do_riscv.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -e
+set -x
+
+# The prebuilt toolchains only support hardfloat, so we only
+# test these for now.
+case $1 in
+  "rv32iv")
+  TARBALL=riscv32-glibc-ubuntu-20.04-nightly-2023.02.25-nightly.tar.gz
```
leekillough, line 10: Is there a reason `2023.02.25` is hard-coded? In `build-riscv.sh`, I use something like this:
```
TAG=$(curl -s https://github.com/riscv-collab/riscv-gnu-toolchain/releases/ \
                  | perl -ne 'if(/(\d\d\d\d\.\d\d\.\d\d)/){print $1; exit}')
```

Then replace all instances of `2023.02.25` with `$TAG`.

leekillough: Even if it must be fixed at `2023.02.25` for now, I would put it in a variable like `TAG` and then use `$TAG` everywhere instead of copying `2023.02.25` in several places.

angsch: Thanks. The main reason for this tag is that it is tested.

kernels/rv64gv/3/bli_gemm_rv64gv_asm_s16x4.c
fgvanzee: No objection here, just want to point out that while `alpha == 0` is handled higher up, `k == 0` must still be handled properly in the microkernel. (This may have gone without saying.)

frame/include/bli_misc_macro_defs.h
```diff
@@ -170,5 +170,7 @@ BLIS_INLINE void bli_toggle_bool( bool* b )
 #define BLIS_VA_END  (-1)
 
 
-#endif
+// Static assertion compatible with any version of C/C++
+#define bli_static_assert(cond) while(0){struct s {int bli_static_assert : !!(cond);};}
```
fgvanzee, line 174: LOL, what is *this*? My brain can't even parse that syntax. 😂

devinamatthews: `!!` = make a conditional into 0 (false) or 1 ( true ). `int whatever : n` this is a bitfield with n>0 bits. Since a bitfield with 0 bits is a compile-time error, then a false condition leads to a compiler error (static assertion failure). One can also try an create and array e.g. `int blis_static_assert[!!(cond)]`. The `while(0)` is for the reason that I've mentioned a coupled of times, where just bare braces could get mis-parsed.

fgvanzee: This is one of those things where I say "live and let live." Weird-ass macro approved! ✅

devinamatthews: I will say it is traditional to name the variable (in the struct) something like 'STATIC_ASSERTION_FAILED' so that people see it in the output and aren't super confused.

leekillough: I tried a dozen syntaxes to try to find a `static_assert` (ala C++11) which is compatible with as many C/C++ dialects as possible, and settled on this one.

The error message is fairly clear, I think.

Let me break down the syntax:
```
while(0) { }
```
This makes it never execute, while still enforcing syntax/semantic checks, and making the whole macro expansion a single compound statement which can be unambiguously embedded inside an `if` statement without braces, similar to the `do { ... } while(0)` paradigm.
```
struct s {int bli_static_assert : !!(cond);};
```
This defines a struct which has a `bli_static_assert` bit-field member whose size depends on `cond`, which must be a compile-time constant (otherwise another error will occur).

If `cond` is true, then `!!(cond)` is `1`, and the bit-field is well-formed.

If `cond` is false, then `!!(cond)` is `0`, and since the bit-field is _named_, it is ill-formed, which causes a compilation error similar to this:
```
foo.c:1:56: error: zero width for bit-field ‘bli_static_assert’
    1 | #define bli_static_assert(cond) while(0){struct s {int bli_static_assert : !!(cond);};}
      |                                                        ^~~~~~~~~~~~~~~~~
foo.c:5:1: note: in expansion of macro ‘bli_static_assert’
    5 | bli_static_assert(0);
      | ^~~~~~~~~~~~~~~~~
```

It also points to the `bli_static_assert()` macro call where it failed.

The bit-field name can be changed to whatever you want it to be. I prefer it is not uppercase unless the macro name itself is uppercase. Since C++11 and C23 have `static_assert` (C11 has `_Static_assert`), I don't think it needs to be all-uppercase.

leekillough: Note: `int blis_static_assert[!!(cond)]` wouldn't work on compilers which support variable-length arrays, because it would allow `cond` to _not_ be a compile-time constant, and even if it was a compile-time constant known to be `0`, some compilers (not many) might even support 0-sized arrays as an extension. Bit fields are better than array sizes in enforcing compile-time constancy and non-zero sizes.

fgvanzee:
@angsch @leekillough The PR was in great shape -- thanks!

Is there anything else you want to throw in before the merge?

frame/include/bli_misc_macro_defs.h
devinamatthews: Re the name I meant `struct s { int STATIC_ASSERT_FAILED : !!(cond)};`, `blis_static_assert` is good.

leekillough:
> Is there anything else you want to throw in before the merge?

Yes, we might add an autodetection header for RISC-V architecture. @angsch and I are still discussing it but it's close to final.

fgvanzee:
@angsch @leekillough Something else that would be useful for me (once you are done with your commits): a final commit log entry that captures everything in this PR. I would normally just use the initial PR description, but I want to give you the chance to augment it in case any other important bits made it in.

If you can provide me with such a draft commit log entry, I'd greatly appreciate it! 🙂

frame/include/bli_misc_macro_defs.h
leekillough: > Re the name I meant `struct s { int STATIC_ASSERT_FAILED : !!(cond)};`, `blis_static_assert` is good.

I'm fine with that. I'll edit it.

kernels/rv64gv/3/bli_gemm_rv64gv_asm_s16x4.c
leekillough: Long-since handled correctly.

angsch:
@fgvanzee @devinamatthews Thanks for your feedback!

Would it be possible to trigger travis-ci manually once on your side? It is not clear why only appveyor has been triggered for the last few commits.

fgvanzee:
> @fgvanzee @devinamatthews Thanks for your feedback!
> 
> Would it be possible to trigger travis-ci manually once on your side? It is not clear why only appveyor has been triggered for the last few commits.

I don't think I can do a manual trigger. But I think I know of a workaround. I can create a local branch at flame/blis, and then you can change the target/destination of this PR to that branch. Then we'll merge it. That will most certainly trigger Travis CI. If something is wrong, it'll be in a branch and give us time to work through the problem. If everything looks fine, then I can merge the local branch into master separately.

How does that sound?

angsch:
> I don't think I can do a manual trigger. But I think I know of a workaround. I can create a local branch at flame/blis, and then you can change the target/destination of this PR to that branch. Then we'll merge it. That will most certainly trigger Travis CI. If something is wrong, it'll be in a branch and give us time to work through the problem. If everything looks fine, then I can merge the local branch into master separately.
How does that sound?

Yes, please! Fingers crossed :-)

fgvanzee:
> Yes, please! Fingers crossed :-)

@angsch I've created a `riscv` branch. Please redirect the PR to that branch, and we'll go from there.

angsch:
> @angsch I've created a riscv branch. Please redirect the PR to that branch, and we'll go from there.

Thanks, new PR is created...

angsch:
Looks like #731 did not trigger Travis either... :-(

leekillough:
Summary of RISC-V changes (with contributions from @angsch):

- The assembly kernels assume the presence of the vector extension `RVV 1.0`.
- It is possible to build `rv[32,64]iv` for any value of `VLEN`. However, if `VLEN < 128`, the targets will fall back to the generic kernels and block sizes.
- The vector microkernels are vector-length agnostic and work with every `VLEN >=128`, but are expected to work best with smaller vector lengths, i.e., `VLEN <= 512`.
- The assembly kernels cover column major storage (`rs_c == 1`).
- The block sizes aim at being a good generic choice for out-of-order cores. They are not tuned to a specific RISC-V HPC core.
- The vector kernels have been tested using `vlen=128, 256, 512`.
- There are four RISC-V base configurations: `rv32i`, `rv32iv`, `rv64i`, and `rv64iv`, namely the 32-bit and 64-bit implementations with and without the `V` vector extension. Additional extensions such as `M` (multiplication), `A` (atomics), `F` (`float` hardware support), `D` (`double` hardware support), and `C` (compressed-length instructions), are automatically used when available. If they are not available, then software equivalents (e.g., `softfloat` and `-latomic`) are used.
- `./configure auto` can be called on a RISC-V build platform, and will automatically detect RISC-V CPU extensions through the [RISC-V C API](https://github.com/riscv-non-isa/riscv-c-api-doc/blob/master/riscv-c-api.md).
- The single- and double-precision assembly code routines for `SGEMM` and `DGEMM`, or for `CGEMM` and `ZGEMM`, are combined in their RISC-V vector assembly source code, and are differentiated only with macros.
- The `XLEN=32` and `XLEN=64` versions of the RISC-V assembly code are identical, except that callee-saved registers are saved and restored differently. There are RISC-V assembly code `#include` files for handling the saving and restoring of callee-saved registers, and they are future-proof if ever `XLEN=128`.
- Multiplications, such as computing array strides and offsets, are performed in C, and later passed to the RISC-V assembly kernels. This is so that the compiler can determine whether the `M` (multiply) extension is available and use multiplication instructions, or call library helper functions instead.
- A new macro called `bli_static_assert()` has been added, to perform static assertions at compile-time, regardless of the C/C++ dialect of the compiler. The original motivation of this was to ensure that calling RISC-V assembly kernels would not silently truncate arguments of type `dim_t` or `stride_t` (so-called "narrowing conversions").
- RISC-V CI tests have been added to Travis, using the `riscv-gnu-toolchain` cross-compiler, and `qemu` simulator.

fgvanzee:
I wonder if there is something going on with the `.travis.yml` file... 🤔 

It's drastic, but we could try squashing/merging and then I could make a trivial change to see if I can get it to trigger?

leekillough:
> I wonder if there is something going on with the `.travis.yml` file... 🤔
> 
> It's drastic, but we could try squashing/merging and then I could make a trivial change to see if I can get it to trigger?

Please do not squash except in the test branch.

fgvanzee:
> Please do not squash except in the test branch.

Just to clarify, you mean the `riscv` branch I recently created for PR #731?

leekillough:
Yes, the temporary branch used to try to trigger CI. You may squash or use other techniques to try to trigger CI, but please do not squash this PR branch.

angsch:
1. To test if this PR breaks `.travis.yml`, we can temporarily remove the CI commit.
2. If there is some quota and I have exhausted it, then removing the CI commit should not make a difference.

If you agree, I will reorder the commits such that the CI commit is last. I will force-push once without the CI commit.

angsch:
Used a rebase to retrigger, CI still not triggered. I will try out Field's idea to squash commits now.

fgvanzee:
How about I take a diff of your entire PR and apply it to a separate branch myself? (We know that branches that I create and push trigger CI, so it will at least allow us to rule out that aspect of it.)

angsch:
> How about I take a diff of your entire PR and apply it to a separate branch myself? (We know that branches that I create and push trigger CI, so it will at least allow us to rule out that aspect of it.)

@fgvanzee  It would be fantastic if you could do this. I did another attempt in #733, where everything is squashed. Also when squashed, Travis is not triggered. In order to rule out that the PR introduced an issue to `.travis.yml`, I removed the 2 RISC-V targets from that file. The result is the same, only Appveyor is triggered.

fgvanzee:
@angsch Okay, I can try that.

Separately, we can investigate something else I just thought of: I'm wondering if you need to be added to the `flame` organization as an external collaborator in order for the CI to trigger.

leekillough:
It was found that Clang, and some gnu toolchain packages, do not automatically enable the `V` extension, and it has to be enabled with `-march=...v`. I have added forcing of `v` in `-march` in https://github.com/angsch/blis/pull/9, which builds on my already-existing RISC-V architecture auto-detection headers.

PR #678, opened 2022-10-23
Omnibus PR - Oct 2023

devinamatthews:
See individual commits for more information. Supersedes #673 and #674.

PR #696, opened 2022-12-10
Fix hobbled multithreaded performance of gemmlike sandbox (and sup when packing)

fgvanzee:
This PR is to track the changes needed to fix a couple (probably related) issues in the recent Omnibus PR #678 (aeb5f0c):
- When configuring with `-s gemmlike` and setting `BLIS_JR_NT=4`, performance of `gemm` is depressed by about 25-30% relative to the preceding commit.
- When configuring with sup enabled and setting `BLIS_JR_NT=4`, performance of the `gemm` sup code path is similarly depressed when optional packing is enabled (via `BLIS_PACK_A=1 BLIS_PACK_B=1`). Performance is unaffected when packing is avoided (the default).

My theory is that every thread is packing every micropanel redundantly, which is an issue that @devinamatthews ran into when developing aeb5f0c. This explanation would be consistent with observations and with the state of the code itself, since the sup `thrinfo_t` tree creation in `bli_l3_thrinfo.c` does not utilize the sub-prenodes like the conventional `thrinfo_t` tree does.

### DIFF
--- a/frame/3/bli_l3_sup_packm.c
+++ b/frame/3/bli_l3_sup_packm.c
@@ -394,7 +394,7 @@ void bli_packm_sup
 			  ( void* )a,  rs_a,  cs_a,
 			          *p, *rs_p, *cs_p,
 			  ( cntx_t* )cntx,
-			  thread
+			  bli_thrinfo_sub_prenode( thread )
 			);
 		}
 		else // if ( schema == BLIS_PACKED_ROW_PANELS )
@@ -415,7 +415,7 @@ void bli_packm_sup
 			          *p, *rs_p, *cs_p,
 			               pd_p, *ps_p,
 			  ( cntx_t* )cntx,
-			  thread
+			  bli_thrinfo_sub_prenode( thread )
 			);
 		}
 
--- a/frame/3/bli_l3_thrinfo.c
+++ b/frame/3/bli_l3_thrinfo.c
@@ -150,6 +150,15 @@ thrinfo_t* bli_l3_sup_thrinfo_create
 	thrinfo_t* thread_jr = bli_thrinfo_split( n_way_jr, thread_pa );
 	thrinfo_t* thread_ir = bli_thrinfo_split( n_way_ir, thread_jr );
 
+	const dim_t n_way_pb = bli_thrinfo_num_threads( thread_pb );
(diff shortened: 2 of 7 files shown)


devinamatthews:
@fgvanzee I pushed a suggested fix to `gemmsup_thread_pack` (sup only, not tested; I don't have a lot of time to work on this right now). You should be able to test and take it from there (gemmlike is very much like sup IIRC).

fgvanzee:
> @fgvanzee I pushed a suggested fix to `gemmsup_thread_pack` (sup only, not tested; I don't have a lot of time to work on this right now). You should be able to test and take it from there (gemmlike is very much like sup IIRC).

Thanks, I'll take a look at your branch and try to pick up where you left off. 👍🏼

fgvanzee:
@devinamatthews Hmm, oddly, your commit definitely helps, but it seems to make up only about half the performance shortfall relative to using the conventional code path.

I'll keep digging into it.

fgvanzee:
@devinamatthews I quickly realized that the true comparison isn't the conventional code path, but rather sup immediately prior to the omnibus commit. Once I did that, your latest code is only about 2% slower. I'll try to see what could account for that, but if I can't find it, I probably won't stress over it.

I'll also look at fixing the sandbox. Thanks for your help.

fgvanzee:
@devinamatthews I quickly realized that the true comparison isn't the conventional code path, but rather sup immediately prior to the omnibus commit. Once I did that, your latest code is only about 2% slower. I'll try to see what could account for that, but if I can't find it, I probably won't stress over it.

I'll also look at fixing the sandbox. Thanks for your help.

> @devinamatthews Hmm, oddly, your commit definitely helps, but it seems to make up only about half the performance shortfall relative to using the conventional code path.
> 
> I'll keep digging into it.

I need more sleep or something. Turns out that *your fix* is 2% faster than the pre-omnibus, not the other way around. Mission accomplished. 🎉

fgvanzee:
> @devinamatthews Hmm, oddly, your commit definitely helps, but it seems to make up only about half the performance shortfall relative to using the conventional code path.
> 
> I'll keep digging into it.

I need more sleep or something. Turns out that *your fix* is 2% faster than the pre-omnibus, not the other way around. Mission accomplished. 🎉

PR #698, opened 2022-12-16
Armv8-A Row-major Kernel Improvements

xrq-phys:
**Status**
This is a 8x6 row-major kernel for ARMv8-A so its internal structure is basically the same as the current 6x8 column-preferring one.

**Updates**
- Instead of clearing **C**-microtile registers at beginning of the assembly, execute the first `k`-loop using `fmul` instead of `fmla`. Codepath within assembly is handled to (basically) not introduce additional branching cost.
- Scatter prefetching code for C into microkernel loops.

**Restrictions**
This kernel assumes hardware prefetching for packed A/B blocks (so as not to bother the pipeline with additional instructions or the DMA with additional loads).
Older chips like ThunderX2 may not perform well with it since they may have *no hardware prefetching* at all, while newer ones like Amazon's C6g tend to be happier with it.

*This update also contains somehow prerequisites for my `gemmsup+packm` work [here](/xrq-phys/blis/tree/packsup).*

### DIFF
--- a/kernels/armv8a/3/armv8a_asm_d2x2.h
+++ b/kernels/armv8a/3/armv8a_asm_d2x2.h
@@ -47,9 +47,25 @@
 " fmla  v"#C0".2d, v"#A".2d, v"#B".d[0] \n\t" \
 " fmla  v"#C1".2d, v"#A".2d, v"#B".d[1] \n\t"
 
+#define DGEMM_2X2_NANOKERNEL_PLAIN(C0,C1,A,B) \
+  DGEMM_2X2_NANOKERNEL(C0,C1,A,B)
+
+#define DGEMM_2X2_NANOKERNEL_INIT(C0,C1,A,B) \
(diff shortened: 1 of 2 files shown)


fgvanzee:
Thanks @xrq-phys! I've asked Jeff to take a look at the new kernel for feedback. I think he and his application could stand to benefit from this, given the inherent advantage row-preferring kernel have with left-sided `trsm` (which is the only `trsm` code path that BLIS implements).

Happy holidays! 🎄 🎁 🍾

GodTamIt:
Hi there, I know this is a bit old but came across this change from [this paper](https://arxiv.org/pdf/2302.08417.pdf).

I was just wondering what the status was for having this (and other changes) merged upstream and/or if there was a plan to do so?

fgvanzee:
Hey @GodTamIt, thanks for your inquiry. I guess we're still waiting on @jdiamondGitHub to look over this PR. I'll reach out to him separately as well.

PR #705, opened 2023-01-04
Disable power10 kernels other than sgemm and dgemm

nisanthmp:
There is a power10 sandbox which uses micro-kernels for (lower precision) datatypes other than float and double. In the generic power10-configured build (non-sandbox), there were compile errors for some microkernels other than sgemm and dgemm. So enabling those kernels only for power10 sandbox (disabling them in the case of normal power10-configured build).


fgvanzee:
This looks good too, @nisanthmp. Thanks again.

PR #706, opened 2023-01-04
Fix k = 0 edge case in power10 micro kernels

nisanthmp:
When sgemm and dgemm microkernels of power10 are called with k = 0, they run into infinite loops and segfault. This is fixed now by early exit in the case of k = 0.


fgvanzee:
Thank you, @nisanthmp! I'll merge this.

PR #707, opened 2023-01-05
Rework bli_packm_blk_var1.

devinamatthews:
Separate the dense code (general/hermitian/symmetric/triangular fully-stored part) from the triangular, diagonal-intersecting code. This allows a more consistent usage of round-robin thread scheduling (even for the dense micro-panels) for the latter.


devinamatthews:
@fgvanzee let me know when you've checked it.

fgvanzee:
@devinamatthews Looks good. Unless there's more you want to do here, I'll squash with a new commit log entry once CI finishes.

PR #709, opened 2023-01-12
Implement `cntx_t` pointer caching in gks.

fgvanzee:
Details:
- Refactored the gks `cntx_t` query functions so that: (1) there is a clearer pattern of similarity between functions that query a native context and those that query its induced (1m) counterpart; and (2) queried `cntx_t` pointers (for both native and induced `cntx_t` pointers) are cached (by default), or deep-queried upon each invocation, depending on whether cpp macro `BLIS_ENABLE_GKS_CACHING` is defined.
- Refactored query-related functions in `bli_arch.c` to cache the queried `arch_t` value (by default), or deep-query the `arch_t` value upon each invocation, depending on whether cpp macro `BLIS_ENABLE_GKS_CACHING` is defined.
- Tweaked the behavior of `bli_gks_query_ind_cntx_impl()` (formerly named `bli_gks_query_ind_cntx()`) so that the induced method `cntx_t` struct is repopulated each time the function is called. (It is still only allocated once on first call.) This was mostly done in preparation for some future in which the `arch_t` value might change at runtime. In such a scenario, the induced method context would need to be recalculated any time the native context changes.
- Added preprocessor logic to `bli_config_macro_defs.h` to handle enabling or disabling of `cntx_t` pointer caching (via `BLIS_ENABLE_GKS_CACHING`).
- For now, `cntx_t` pointer caching is enabled by default and does not correspond to any official `configure` option. Disabling can be done by inserting a `#define` for `BLIS_DISABLE_GKS_CACHING` into the appropriate `bli_family_*.h` header file within the configuration of interest.
- Thanks to Harihara Sudhan S (AMD) for suggesting that `cntxt_t` pointers (and not just `arch_t` values) be cached.
- Comment updates.

@ihariharasudhan Please review these changes. Thanks.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -104,6 +104,7 @@ but many others have contributed code and feedback, including
   Paul Springer            @springer13         (RWTH Aachen University)
   Adam J. Stewart          @adamjstewart       (University of Illinois at Urbana-Champaign)
   Vladimir Sukarev
+  Harihara Sudhan S        @ihariharasudhan    (AMD)
   Chengguo Sun             @chengguosun
   Santanu Thangaraj                            (AMD)
   Nicholai Tukanov         @nicholaiTukanov    (The University of Texas at Austin)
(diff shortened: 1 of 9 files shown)


ihariharasudhan (commented):
Looks good to me.

PR #711, opened 2023-01-13
Merge the tlb-specific gemm macro-kernel with the standard one.

devinamatthews:
These two version can be merged by incorporating iteration information from both approaches: start/end/increment from the standard slab or rr partitioning and number of microtiles + starting location from the tlb partitioning. The slab/rr partitioning simply pretends to be a tlb partitioning which starts each iteration of the ir loop in the same place rather than going back to iteration 0, while conversely the tlb iteration respects the end and increment parameters from the slab/rr partitioning.


fgvanzee:
@devinamatthews Let me know if you have any comments on my changes. Otherwise, I'll start writing a commit log entry and squash.

devinamatthews:
👍

PR #713, opened 2023-01-17
Fixing macro-mismatch error for VEXTRACTF64X2

HarshDave12:
Details:
    - This commit fixes a mismatch between macro mapping.
      Macro VEXTRACTF64X2 was defined as vextractf64x4, instead of
      vextractf64x2.


fgvanzee:
Thank you for catching this, @HarshDave12! I'll add you to the `CREDITS` file and then merge.

PR #714, opened 2023-01-18
Change configure script to use here-document for improved readability and maintainability

leekillough:
Change BLIS `configure` script to use [here-document](https://en.wikipedia.org/wiki/Here_document), for improved readability and maintainability.

It avoids hundreds of `echo` commands. It also avoids having to escape `"` characters in the message. And the source text isn't indented (although it could be indented with `<<-` syntax and _tab_ characters, but not with spaces). It is WYSIWYG.
```
./configure --help > configure.old.out
./configure --help > configure.new.out
diff -s -b configure.old.out configure.new.out
Files configure.old.out and configure.new.out are identical
```
The only difference in the output is that a leading space character at the beginning of each _blank_ line is removed (it was caused by `echo " "`).

### DIFF
--- a/configure
+++ b/configure
@@ -46,7 +46,7 @@ print_usage()
 	fi
 
 	# Echo usage info.
-        cat <<EOF
+ 	cat <<EOF
 
  ${script_name} (BLIS ${version})
 
@@ -436,6 +436,7 @@ print_usage()
    configuration.
 
 EOF
+
 	# Exit with non-zero exit status
 	exit 1
 }
--- a/configure
+++ b/configure
@@ -46,397 +46,396 @@ print_usage()
 	fi
 
 	# Echo usage info.
-	echo " "
-	echo " ${script_name} (BLIS ${version})"
-	#echo " "
-	#echo " BLIS ${version}"
-	echo " "
-	echo " Configure BLIS's build system for compilation using a specified"
-	echo " configuration directory."
-	echo " "
-	echo " Usage:"
-	echo " "
-	echo "   ${script_name} [options] [env. vars.] confname"
-	echo " "
-	echo " Arguments:"
-	echo " "
-	echo "   confname      The name of the sub-directory inside of the 'config'"
-	echo "                 directory containing the desired BLIS configuration."
-	echo "                 Note that confname MUST be specified; if it is not,"
-	echo "                 configure will complain. To build a completely generic"
-	echo "                 implementation, use the 'generic' configuration"
-	echo " "
-	echo " Options:"
-	echo " "
-	echo "   -p PREFIX, --prefix=PREFIX"
-	echo " "
-	echo "                 The common installation prefix for all files. If given,"
-	echo "                 this option effectively implies:"
-	echo "                   --libdir=EXECPREFIX/lib"
-	echo "                   --includedir=PREFIX/include"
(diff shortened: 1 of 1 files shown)


fgvanzee:
Thanks, @leekillough! This is a nice improvement. (I didn't know about here-documents prior to this... TIL.) It will make it *much* easier to manage `configure` options in the future. I'll merge it after CI finishes (a formality, granted). Appreciate your contribution.

leekillough:
I'm surprised the small whitespace changes you made and the beginning and end made GitHub compare suddenly line the text up (previously there were two separate large blobs).

I untabified the source lines and text output of `configure --help`. I try not to make assumptions about tab spacing on terminals, so to line things up, at least on fixed-width fonts, I use spaces instead of tabs in text output. I care less about tabs vs. spaces in source code, although I lean towards spaces because they are more uniform across different platforms, editors, etc. I preserved the use of tabs in `emacs c-mode` in the `.dir-locals.el` because that was what was there before, and because tabs are mentioned in the contributor guidelines.

fgvanzee:
> I'm surprised the small whitespace changes you made at the beginning and end made GitHub compare suddenly line the text up (previously there were two separate large blobs).

I still see the two large blobs. Maybe I'm missing something.

In BLIS, we use both tabs and spaces in a way that provides a malleable (and yet also predictably tidy) formatting style regardless of the user's tab width. But I don't expect others to perfectly use that format in their PRs, so I'm always happy to make those tweaks myself.

Thanks again for this change. I'm ready to squash if you're happy with it.

leekillough:
> I still see the two large blobs. Maybe I'm missing something.

If you use `git diff` locally it might be different (sometimes using `-b` or even `-w` may avoid extraneous diffs). I was referring to the GitHub web interface's `Files Changed` tab, which used to have two large blobs separated from each other, but with your small change, now they almost line up perfectly.

>  I'm willing to make an exception to BLIS's typical tabbing conventions due to the WYSIWYG here-document syntax. I understand that it's a literal string, and I can get by without any tabs/indenting for that block of code.

That's what I figured -- make the literal text output use only spaces, without changing the rest of the script.

> Thanks again for this change. I'm ready to squash if you're happy with it.

Sure, go ahead.

fgvanzee:
> I was referring to the GitHub web interface's `Files Changed` tab, which used to have two large blobs separated from each other, but with your small change, now they almost line up perfectly.

Strange. I am also looking at the `Files Changed` tab. Are you using Unified or Split view? I tried both, with and without whitespace hidden. In Unified, it always showed up as two big blobs. In Split, they were almost lined up, but still lots of red and green. 🤔

leekillough:
I'm using split. I saw two large separated blobs before your change, with no red or green lines in common: long red lines with nothing to the right of it, followed by long green lines with nothing to the left of it. After your minor whitespace change, the red and green mostly lined up.

leekillough:
See the difference here: https://github.com/flame/blis/pull/714/commits/b69375038a3e87680192f49764ca9174bf59d356

fgvanzee:
> See the difference here: [b693750](https://github.com/flame/blis/commit/b69375038a3e87680192f49764ca9174bf59d356)

Ahhh, yes, okay. I can see how the final diff was a vast improvement!

PR #716, opened 2023-01-26
Add casts to printf arguments to prevent compiler warnings

leekillough:
When building 32-bit targets such as RISC-V `rv32iv`, `dim_t` and other types are declared as `int`.

This produces warnings of `printf` argument type mismatch in `bli_thread_range_tlb.c`, although most of the time, these warnings are spurious, because `int == long`.

This change removes those warnings, by explicitly casting `printf` arguments to type `long`.

Cc: @angsch @devinamatthews

### DIFF
--- a/frame/thread/bli_thread_range_tlb.c
+++ b/frame/thread/bli_thread_range_tlb.c
@@ -142,11 +142,11 @@ dim_t bli_thread_range_tlb_l
 	const dim_t offn_ut_nonrect = ( diagoffmin / nr );
 
 	PGUARD printf( "---------------------------\n" );
-	PGUARD printf( "min(diagoff,n):     %7ld\n", diagoffmin );
-	PGUARD printf( "offn_ut_nonrect:    %7ld\n", offn_ut_nonrect );
-	PGUARD printf( "offn_nonrect:       %7ld\n", offn_nonrect );
-	PGUARD printf( "diagoff_nonrect:    %7ld\n", diagoff_nonrect );
-	PGUARD printf( "n_nonrect:          %7ld\n", n_nonrect );
+	PGUARD printf( "min(diagoff,n):     %7ld\n", (long) diagoffmin );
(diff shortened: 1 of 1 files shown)


fgvanzee:
@leekillough I wanted to clean up some whitespace, but it appears I don't have permissions to push to this branch. Is that something you can fix on your end?

leekillough:
I added you as a collaborator on my repo.I don't care about whitespace, but I tend to prefer spaces to tabs, since they are more uniform.In emacs it's easy to "tabify" or "untabify".LeeOn Jan 27, 2023, at 12:23, Field G. Van Zee ***@***.***> wrote:﻿
@leekillough I wanted to clean up some whitespace, but it appears I don't have permissions to push to this branch. Is that something you can fix on your end?

fgvanzee:
@leekillough I understand if you prefer to keep your branch formatted the way it is. I can fix the whitespace in `flame:master` after the merge.

leekillough:
I don't care about whitespace; feel free to format it however you want.

The line breaks I added, were only added when the (long) casts made the lines go longer than 80 characters when they weren't already longer than 80 characters.

leekillough:
Again, it's no big deal. Feel free to format it however you want to. I was just contributing to remove the printf warnings.

If you want me to manually reformat according to some criteria before merge, I can certainly do that.

fgvanzee:
Thanks @leekillough. My commit also adds a few `long` typecasts that were missing from the original commit.

leekillough:
Sorry if I missed some. I tried to be thorough.

fgvanzee:
It's no problem, Lee. This is why we do reviews. 🙂 I'm used to it. (And yes, you were quite thorough.)

leekillough:
I only want the best for BLIS :-)

fgvanzee:
We appreciate your contributions! 🚀

PR #718, opened 2023-02-01
add runtime selection of power architecture

nisanthmp:
use "./configure power" and "make" to create a single shared library for power10/power9/generic.
With this feature, blis shared library could be compiled on a power9 system and run on a power10 system and vice-versa. If run on a generic system, unoptimised generic code will execute.

This will work with gcc only, as that is the only compiler supported by both power9 and power10 configurations in BLIS

### DIFF
--- a/config/power/bli_family_power.h
+++ b/config/power/bli_family_power.h
@@ -1,0 +1,41 @@
+/*
+
+   BLIS
+   An object-based framework for developing high-performance BLAS-like
+   libraries.
+
+   Copyright (C) 2014, The University of Texas at Austin
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions are
+   met:
+    - Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    - Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    - Neither the name(s) of the copyright holder(s) nor the names of its
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+   "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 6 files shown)


nisanthmp:
This PR might help solve this issue: https://github.com/flame/blis/issues/691

devinamatthews:
Sorry, I mean #345.

nisanthmp:
I have tested on Power9 and Power10.
The method used is reading the proc/cpuinfo file

devinamatthews:
Ah, I did not realize that PPC had gotten added in to the ARM32 cpuid section.

devinamatthews:
OK, the changes look good, but could you please update the documentation in `configure` and in `docs`?

nisanthmp:
@devinamatthews 

configure does not have all the compiler version checks for IBM systems - but it is wrto individual architectures (power9, power10 etc), rather than the family names (power) which is the subject of this PR. I could not find any other place where the configure (including documentation) changes related to this PR - can you please help me figure out?

same with docs, individual architectures (power9, power10 etc) need to be updated with support, but there is nothing wrto configuration families (as the currently added configuration - ‘power’ is a umbrella family for power9 and power10), that I could find that needs to be updated. So my opinion is that they should go in a separate issue of BLIS and/or a separate PR.  

Please do help me figure out if I missed something…  

Please note that fixing the individual architecture documentation for compiler version checks etc for IBM Power architectures is a bit tricky as of now, as even the latest generally distributed gcc does not build all code written for power10 - you need RH distributed gcc with power10 tool chain support for that. So just updating the gcc version checks will blacklist power10 completely.

fgvanzee:
@nisanthmp Was your previous comment stemming from trying to interpret Devin's request to update the documentation in `configure` and `docs`? If so, I wasn't quite sure what he meant by that, either.)

devinamatthews:
Devin is having difficultly responding to things recently so sorry. I meant that I thought there were a few places where supported configurations (including configuration families) were listed in the docs and/or in output (--help) from configure. If that's not the case then please merge away.

fgvanzee:
> Devin is having difficultly responding to things recently so sorry. I meant that I thought there were a few places where supported configurations (including configuration families) were listed in the docs and/or in output (--help) from configure. If that's not the case then please merge away.

Ah, gotcha. `configure` actually does an `ls` to get the list of currently supported configurations (when a config target is not given, for example). We *do* have a document that lists our current architecture support, but I always forget to update it. (If @nisanthmp wants to update that doc, he is welcome to do so.)

Hopefully that clears up some confusion, @nisanthmp. You may still have questions; if so, please let us know.

nisanthmp:
@fgvanzee and @devinamatthews - Added the missing documentation in hardware support doc now. Please have a look.

fgvanzee:
I'm fine with this. Thanks for your patience, @nisanthmp.

PR #719, opened 2023-02-03
add nvhpc compiler support

ajaypanyala:


config/generic/make_defs.mk
```diff
@@ -70,8 +70,11 @@ CKVECFLAGS     :=
 else
 ifeq ($(CC_VENDOR),clang)
 CKVECFLAGS     :=
+ifeq ($(CC_VENDOR),nvc)
```
devinamatthews, line 73: There is a logic bug somewhere here which is breaking builds with `clang`.

ajaypanyala: Sorry for the delay, it's fixed now.

PR #720, opened 2023-02-05
Define `BLIS_VERSION_STRING` in `blis.h`.

fgvanzee:
Details:
- Previously, the version string was communicated from `configure` to `config.mk` (via the `config.mk.in` template), where it was included via the top-level `Makefile`, where it was then used to define the preprocessor macro `BLIS_VERSION_STRING` via a command line argument to the compiler (via `-D`). This macro is then used within `bli_info.c` to initialize a static string which can then be queried via the `bli_info_get_version_str()` function. However, there are some applications that may find utility in being able to access the version string without actually linking against BLIS. This commit moves the definition of `BLIS_VERSION_STRING` into `bli_config.h` (via the `bli_config.h.in` template) so that it is available in the monolithic (flattened) `blis.h` that is created at compile-time and installed alongside of the library. (Note that `bli_info_get_version_str()` is still available at runtime; the only thing that changed was where/how the string constant was defined.) Thanks to Mohsen Aznaveh and Tim Davis for providing the idea for this change.

cc: @Aznaveh

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -3,126 +3,128 @@ BLIS framework
 Acknowledgements
 ---
 
-The BLIS framework was primarily authored by
+The BLIS framework was originally authored by
 
-  Field Van Zee            @fgvanzee           (The University of Texas at Austin)
+  Field Van Zee            @fgvanzee                  (The University of Texas at Austin)
 
-but many others have contributed code and feedback, including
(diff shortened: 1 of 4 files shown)


fgvanzee:
See any issues with this, @devinamatthews?

devinamatthews:
Seems fine to me.

PR #722, opened 2023-02-10
Use 'const' pointers in kernel APIs.

fgvanzee:
Details:
- Qualified all input-only data pointers in the various kernel APIs with the `const` keyword while also removing `restrict` from those kernel APIs. (Use of `restrict` was maintained in kernel implementations, where appropriate.) This affected the function pointer types defined for all of the kernels, their prototypes, and the reference and optimized kernel definitions' signatures.
- Templatized the definitions of `copys_mxn` and `xpbys_mxn` static inline functions.
- Minor whitespace and style changes (e.g. combining local variable declaration and initialization into a single statement).
- Removed some unused kernel code left in `old` directories.
- In `common.mk`, `grep` for `"#include"` instead of `"\#include"`. Not sure why the backslash is suddenly giving a `"grep: warning: stray \ before #"` message, but removing the backslash seems to fix the issue.

@devinamatthews Note that this does *not* yet contain the switchover to `void*`. I wanted to do it in two phases in case anything went wrong.


leekillough:
If I may respectfully ask, what is the reason for removing `restrict`, and why wasn't `const` already used for pointers to read-only data?

devinamatthews:
Re `restrict`, it's not *really* needed in the function prototypes or typedefs, since it only has an effect on the generated code when used in the definition. True, one could argue it serves a purpose as defining a user contract, but it's not quite nuanced enough to do that precisely for the kernels anyways. It is still used when writing kernel definitions usually.

Re `const`, the codebase didn't use `const` at all until I started putting it in a year ago or so. This is just an evolution of those efforts.

leekillough:
A caller is not affected by knowing whether two declared pointer parameters are aliased, unless one of them is read-only (pointer to `const`) and the other is read-write, and the caller reads the read-only one before and after the call, and because it knows of the `restrict` in the parameter declaration, it can assume that the read-only one was not affected during the call. But that occurs in a very small number of real cases.

I am glad you're updating the appropriate use of `const` where it wasn't used before.

devinamatthews:
> A caller is not affected by knowing whether two declared pointer parameters are aliased (or declared restrict), unless one of them is read-only (pointer to const) and the other is read-write, and the caller reads the read-only one before and after the call, and because it knows of the restrict in the parameter declaration, it can assume that the read-only one was not affected during the call. But that only occurs in a very small number of real cases.

Well, `restrict` on an interface is only a pinky-swear to the compiler. The implementation can ignore that restriction (same with const).

leekillough:
Also, the caller can only assume that the `const` `* restrict` addressed parameter was not modified if its address was not "exposed" globally or through some other less restrictive call elsewhere (so-called "address exposed"). So the benefit of `restrict` is highly `restrict`ed.

Except in the innermost kernels, I agree that the utility of `restrict` is limited.

fgvanzee:
For anyone not following the Discord discussion, these CI tests are failing because I removed the `\` from:
```
REF_KER_HEADERS := $(shell $(GREP) "\#include" $(REF_KER_SRC) | ... )
```
in `common.mk` because as of grep 3.8, that backslash is causing the following warning:
```
grep: warning: stray \ before #" message
```
The warning is obviously undesired, but even more undesired is the build system breaking altogether. So I'm trying to figure out a way to satisfy both pre- and post-3.8 greps.

leekillough:
I believe that `#` must be escaped because otherwise gmake assumes it's the start of the comment and ignores the rest of the line, obviously leading to a syntax error.

fgvanzee:
> I believe that `#` must be escaped because otherwise gmake assumes it's the start of the comment and ignores the rest of the line, obviously leading to a syntax error.

That kind of makes sense, but GNU `make` works as expected with:
```
REF_KER_HEADERS := $(shell $(GREP) "#include" $(REF_KER_SRC) | ... )
```
when using grep 3.8. So... I dunno.

fgvanzee:
While I don't like it, the only band-aid I could think of is to redirect `stderr` to `/dev/null` for that particular command sequence. (See #723.)

fgvanzee:
@nisanthmp Please try compiling the power kernels in this commit to make sure I didn't break anything. If you get any errors (or warnings), please report them here so I can fix them.

fgvanzee:
Thank you, @nisanthmp. I think I've fixed the errors that you reported. Could you please confirm there are no others?

nisanthmp:
@fgvanzee There are no other compilation errors when configured for power family, ie power9 and power10 configurations

fgvanzee:
> @fgvanzee There are no other compilation errors when configured for power family, ie power9 and power10 configurations

Wonderful. Thanks for your help!

angsch:
Cosmetics: Some of the non-reference kernels received bonus `\`. For example:

https://github.com/flame/blis/blob/93c63d1f469c4650df082d0fa2f29c46db0e25f5/kernels/armsve/3/bli_gemm_armsve_asm_d2vx10_unindexed.c#L46-L57

PR #727, opened 2023-02-20
Use `void*` datatypes in kernel APIs.

fgvanzee:
Details:
- Migrated all kernel APIs to use `void*` pointers instead of `float*`, `double*`, `scomplex*`, and `dcomplex*` pointers. This allows us to define many fewer kernel function pointer types, which also makes it much easier to know which function pointer type to use at any given time. (For example, whereas before there was `?axpyv_ker_ft`, `?axpyv_ker_vft`, and `axpyv_ker_vft`, now there is just `axpyv_ker_ft`, which is equivalent so what `axpyv_ker_vft` used to be.)
- Refactored how kernel function prototypes and kernel function types are defined so as to reduce redundant code. Specifically, the function signatures (excluding `cntx_t*` and, in the case of level-3 microkernels, `auxinfo_t*`) are defined in new headers named, for example, `bli_l1v_ker_params.h`. Those signatures are reused via macro instantiation when defining both kernel prototypes and kernel function types. This will hopefully make it a little easier to update, add, and manage kernel APIs going forward.
- Updated all reference kernels according to the aforementioned switch to `void*` pointers.
- Updated all optimized kernels according to the aforementioned switch to `void*` pointers. This sometimes required renaming variables, inserting typecasting so that pointer arithmetic could continue to function as intended, and related tweaks.
- Updated `sandbox/gemmlike` according to the aforementioned switch to `void*` pointers.
- Renamed (to better align with naming of neighboring files):
  - `frame/1/bli_l1v_ft_ker.h    -> frame/1/bli_l1v_ker_ft.h`
  - `frame/1f/bli_l1f_ft_ker.h   -> frame/1f/bli_l1f_ker_ft.h`
  - `frame/1m/bli_l1m_ft_ker.h   -> frame/1m/bli_l1m_ker_ft.h`
  - `frame/3/bli_l1m_ft_ukr.h    -> frame/3/bli_l1m_ukr_ft.h`
  - `frame/3/bli_l3_sup_ft_ker.h -> frame/3/bli_l3_sup_ker_ft.h`
- Added the missing `void* params` argument to `bli_?packm_struc_cxk()` in `frame/1m/packm/bli_packm_struc_cxk.c`. This argument is being passed into the function from `bli_packm_blk_var1()`, but wasn't being "caught" by the function definition itself. The function prototype for `bli_?packm_struc_cxk()` also needed updating.
- Reordered the last two parameters in `bli_?packm_struc_cxk()`. (Previously, the `void* params` was passed in after the `const cntx_t* cntx`, although because of the above bug the `params` argument wasn't actually present in the function definition.)

cc: @devinamatthews Here's phase 2 of #722.


fgvanzee:
@nisanthmp Like you did with #722, could you please test out the changes in this commit on the power microarchitectures? Thanks so much.

fgvanzee:
@xrq-phys Would you mind similarly reviewing/testing the arm kernels? Travis CI does some checks already, but given the extent of the changes a human touch would be appreciated. 😄

xrq-phys (approved):
No problem seem to exist in code itself.

Running *one* additional test on NEON SUP.

xrq-phys:
Check passed. :D

nisanthmp:
@fgvanzee compiles and checks fine on power (power9 and power10) :-)

fgvanzee:
> @fgvanzee compiles and checks fine on power (power9 and power10) :-)

Glad to hear. Thank you for your assistance!

jeffhammond:
Or you could just use C++ internally, while not using any C++ features that would require the runtime library, and not have to play games with void pointers.

devinamatthews:
@jeffhammond you're preaching to the choir but there's apparently not a C++ toolchain at all on enough platforms to make a difference. How this is possible I don't know.

jeffhammond:
I've programmed some really crazy even imaginary hardware in my career and the C++ required for this worked on all of it, even with the IBM XL C++ compiler on 32-bit Blue Gene. 

I would argue that the vendors who build garbage that can handle this should pay you for the maintenance cost associated with their stupidity.

It isn't my area but I think people use the aforementioned C++ subset on microcontrollers and other lightweight platforms.

leekillough:
@jeffhammond : LOL

PR #729, opened 2023-03-16
Make the BLIS configure script pass all shellcheck checks

leekillough:
Make the BLIS `configure` script pass all `shellcheck` checks, disabling ones which we violate but which are just stylistic, or are special cases in our code.

1. `# shellcheck disable=2001,2249,2034,2154,2181,2312,2250,2292` disables the following checks for the entire `configure` script:

- [SC2001](https://www.shellcheck.net/wiki/SC2001): See if you can use `${variable//search/replace}` instead.
- [SC2249](https://www.shellcheck.net/wiki/SC2249): Consider adding a default `*)` case, even if it just exits with error.
- [SC2034](https://www.shellcheck.net/wiki/SC2034): `foo` appears unused. Verify it or export it.
- [SC2154](https://www.shellcheck.net/wiki/SC2154): `var` is referenced but not assigned. (`configure` uses `eval` to assign to variables whose names are runtime variables.)
- [SC2181](https://www.shellcheck.net/wiki/SC2181): Check exit code directly with e.g. `if mycmd;`, not indirectly with `$?`.
- [SC2312](https://www.shellcheck.net/wiki/SC2312): Consider invoking this command separately to avoid masking its return value (or use `|| true` to ignore).
- [SC2250](https://www.shellcheck.net/wiki/SC2250): Prefer putting braces around variable references even when not strictly required.
- [SC2292](https://www.shellcheck.net/wiki/SC2292): Prefer `[[ ]]` over `[ ]` for tests in Bash/Ksh.

2. A bug was found in `pass_config_kernel_registries()` which prevented a list of more than one configuration from working:
```
        for item in "${list}"; do
```
The `"${list}"` is taken as one item, and word splitting is not performed, so the loop only executes once. It should be:
```
        for item in ${list}; do
```
or it should use Bash array variables.

3. A bug was found in `main()`:

```
       while $found = true; do
```
should be:
```
       while [[ $found = true ]]; do
```

4. Address [SC2268](https://www.shellcheck.net/wiki/SC2268): Avoid x-prefix in comparisons as it no longer serves a purpose. (e.g., `[ "x$var" = "xvalue"]`).

5. Address [SC2086](https://www.shellcheck.net/wiki/SC2086): Double quote to prevent globbing and word splitting.

There are many places in the code where variables are expanded but are not quoted, which causes word splitting, and globbing if the variable contains wildcard characters, and can lead to syntax errors if the variable is empty and used in `[ ]`.

When the variable (or command substitution `$( ... )`) appears on the LHS or as a unary operand of a `[ ]` test, then simply replacing `[ ]` with `[[ ]]` will avoid the problem, because inside `[[ ]]`, the expansion is syntactically always present in the test, even if it expands to empty.

When a variable possibly contains spaces, and it needs to be passed or treated as a single value, it should be double-quoted. As a test, I tried building BLIS in a directory whose name contained spaces.

On the other hand, there are a few places where word splitting is desired, such as building command lines with multiple options separated by spaces, and for those cases, this check has been disabled with `# shellcheck disable=2086`.

6. As part of <span>#</span>5, the variables were placed inside double quotes to avoid [SC2086](https://www.shellcheck.net/wiki/SC2086). But I noticed that when printing the final installation directories, `configure` was sometimes printing symbolic names like `${prefix}`. To fix this, I added `eval echo`, two for `libdir` in case it defaults to `${exec_prefix}` and `exec_prefix` defaults to `${prefix}`:

```
       echo "${script_name}:   prefix:      "${prefix}
       echo "${script_name}:   exec_prefix: "${exec_prefix}
       echo "${script_name}:   libdir:      "${libdir}
       echo "${script_name}:   includedir:  "${includedir}
       echo "${script_name}:   sharedir:    "${sharedir}
```
becomes:
```
       echo "${script_name}:   prefix:      $(eval echo "${prefix}")"
       echo "${script_name}:   exec_prefix: $(eval echo "${exec_prefix}")"
       echo "${script_name}:   libdir:      $(eval echo "$(eval echo "${libdir}")")"
       echo "${script_name}:   includedir:  $(eval echo "${includedir}")"
       echo "${script_name}:   sharedir:    $(eval echo "${sharedir}")"
```
It now prints out like this:
```
configure: final installation directories:
configure:   prefix:      /usr/local
configure:   exec_prefix: /usr/local
configure:   libdir:      /usr/local/lib
configure:   includedir:  /usr/local/include
configure:   sharedir:    /usr/local/share
```

7. Address [SC2162](https://www.shellcheck.net/wiki/SC2162): `read` without `-r` will mangle backslashes.

There are a few places where `read` is used without `-r`. Unless it is the intention to interpret backslashes as escape characters in files read by `configure`, `read -r` should be used.

8. Address [SC2196](https://www.shellcheck.net/wiki/SC2196): `egrep` is non-standard and deprecated. Use `grep -E` instead.

9. Address [SC2244](https://www.shellcheck.net/wiki/SC2244): Prefer explicit `-n` to check non-empty string (or use `=`/`-ne` to check boolean/integer).

10. Address [SC2295](https://www.shellcheck.net/wiki/SC2295): Expansions inside `${..}` need to be quoted separately, otherwise they will match as a pattern. ([Unclear](https://github.com/koalaman/shellcheck/issues/2385)?) 

```
       dist_path=${0%/${script_name}}
```
```
       dist_path=${0%"/${script_name}"}
```

11. Address [SC2004](https://www.shellcheck.net/wiki/SC2004): `$/${}` is unnecessary on arithmetic variables.
```
       shift $(($OPTIND - 1))
```
becomes:
```
       shift $((OPTIND - 1))
```
12. Address [SC2006](https://www.shellcheck.net/wiki/SC2006): Use  `$(...)` notation instead of legacy backticked `` `...` ``.
```
                                       var=`expr "$1" : '\([^=]*\)='`
                                       value=`expr "$1" : '[^=]*=\(.*\)'`
                                       eval $var=\$value
                                       export $var
```
becomes:
```
                                       var=$(expr "$1" : '\([^=]*\)=')
                                       value=$(expr "$1" : '[^=]*=\(.*\)')
                                       eval "export $var=\$value"
```

13. Address [SC2002](https://www.shellcheck.net/wiki/SC2002): Useless `cat`. Consider `cmd < file`.

```
       so_version_major=$(cat ${so_version_filepath} | sed -n "1p")
```
becomes:
```
       so_version_major=$(sed -n "1p" < "${so_version_filepath}")
```
and:

```
       so_version_minorbuild=$(cat ${so_version_filepath} | sed -n "2p")
```
becomes:
```
       so_version_minorbuild=$(sed -n "2p" < "${so_version_filepath}")
```
and others.

14. Address [SC2166](https://www.shellcheck.net/wiki/SC2166): Prefer `[ p ] && [ q ]` as `[ p -a q ]` is not well defined.

Also, for style, use `[[ ]]` which has `&&` and `||` operators, instead of `[ ] && [ ]` and `[ ] || [ ]`.

Replace:
```
       if [ "${cc_vendor}" = "icc" -o \
            "${cc_vendor}" = "gcc" ]; then
```
with:
```
       if [[ ${cc_vendor} = icc || ${cc_vendor} = gcc ]]; then
```
Similarly, replace:
```
               if   [ "${env_str}" = "AR" -o \
                      "${env_str}" = "RANLIB" ]; then
```
with:
```
               if   [[ ${env_str} = AR || ${env_str} = RANLIB ]]; then
```
17. (Style.) Replace:
```
               if [ "$(echo ${vendor_string} | grep -o Apple)" = "Apple" ]; then
```
with:
```
               if [[ ${vendor_string} = *Apple* ]]; then
```
18. (Style.) Replace:
```
               if [[ "${list}" == *[/]* ]]; then
```
with:
```
               if [[ ${list} = */* ]]; then
```

19. (Style.) Replace long pipelines of `sed` and `perl` filters with multiple `-e` scriptlets. Faster and creates fewer processes. Need to add `;` at the end of each Perl scriptlet, since they get concatenated into one large script.

20. (Style.) Replace 7 long `${gen_make_frags_sh}` calls, with all but 3 arguments the same in each case, and with an `echo` before each call, with short calls to a new`create_makefile_fragment()` function.

21. (Style.) Replace repeated code which creates `Makefile`, `blis.pc.in`, `common.mk` and `config` symlinks, with one loop which iterates over them. Also address [SC2226](https://www.shellcheck.net/wiki/SC2226): This `ln` has no destination. Check the arguments, or specify `.` explicitly.

22. (Style.) Rather than setting a variable with a `grep` command substitution and then testing whether it is empty, test the assignment command itself, which will return a nonzero exit code if the `grep` fails to find a match:
```
       aocc_grep=$(echo "${vendor_string}" | grep 'AOCC')
       if [ -n "${aocc_grep}" ]; then
```
becomes:
```
       if aocc_grep=$(echo "${vendor_string}" | grep 'AOCC'); then
```
and:
```
               aocc_ver21=$(echo "${vendor_string}" | grep 'AOCC.LLVM.2')
               if [ -n "${aocc_ver21}" ]; then
```
becomes:
```
               if aocc_ver21=$(echo "${vendor_string}" | grep 'AOCC.LLVM.2'); then
```
This allows the program to continue in case `set -e` is enabled, because it's not considered a fatal error if the `grep` fails inside an `if`.

23.  (Style.) Use `bash` pattern-matching instead of `echo` and `grep` (supersedes <span>#</span>22): 
```
               uconf=$(echo "${config_name}" | grep -c 'zen\|amd64')

               if [[ $uconf == 0 ]]; then
```
becomes:
```
               if [[ ${config_name} != *zen* && ${config_name} != *amd64* ]]; then
```
and:
```
       aocc_grep=$(echo "${vendor_string}" | grep 'AOCC')
       if [ -n "${aocc_grep}" ]; then
```
becomes:
```
       if [[ ${vendor_string} = *AOCC* ]]; then
```
and:
```
       armclang_grep=$(echo "${vendor_string}" | grep 'Arm C/C++/Fortran Compiler')
       if [ -n "${armclang_grep}" ]; then
```
becomes:
```
       if [[ ${vendor_string} = *'Arm C/C++/Fortran Compiler'* ]]; then
```

24. Make the `configure` script work when errors are fatal (`set -e`). Add `|| :` in a few places, such as:
```
                       vendor_string="$(${FC} --version 2>/dev/null || :)"
```
this needs to be tested on all supported platforms, perhaps in the CI, to ensure that `set -e` can been enabled in `configure`.

**Edit:** In case it wasn't clear, `set -e` has not been added to the `configure` script, since it may break it on certain platforms I cannot test; however, clearly obvious places where `set -e` would cause `configure` to fail have been identified and remedied. With the `|| :` change above, `configure` works fine to completion on x86-64 even if `set -e` is added before `main` is called. I consider it good practice to make sure a script properly handles every command which returns a non-zero exit code.

25. (Style.) Use `$(<file)` instead of `$(cat file)` [for loading the contents of a file into a shell variable](https://github.com/koalaman/shellcheck/issues/2493).

### DIFF
--- a/configure
+++ b/configure
@@ -32,6 +32,7 @@
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 #
+# shellcheck disable=2001,2249,2034,2154,2181,2312,2250,2292
 
 #
 # -- Helper functions ----------------------------------------------------------
@@ -46,7 +47,7 @@ print_usage()
 	fi
 
 	# Echo usage info.
- 	cat <<EOF
+	cat <<EOF
 
  ${script_name} (BLIS ${version})
 
@@ -562,7 +563,7 @@ pass_config_kernel_registries()
 
 		# We've stripped out leading whitespace and trailing comments. If
 		# the line is now empty, then we can skip it altogether.
-		if [ "x${curline}" = "x" ]; then
+		if [[ -z ${curline} ]]; then
 			continue;
 		fi
 
@@ -572,12 +573,13 @@ pass_config_kernel_registries()
 
 		# If we encounter a slash, it means the name of the configuration
 		# and the kernel set needed by that configuration are different.
-		if [[ "${list}" == *[/]* ]]; then
+		if [[ ${list} = */* ]]; then
 
 			#echo "Slash found."
 			klist=""
 			clist=""
-			for item in "${list}"; do
+
+			for item in ${list}; do
 
 				# The sub-configuration name is always the first sub-word in
 				# the slash-separated compound word.
@@ -590,7 +592,7 @@ pass_config_kernel_registries()
 
 				# Replace the slashes with spaces to transform the string
 				# into a space-separated list of kernel names.
-				kernels=$(echo -e ${kernels} | sed -e "s/\// /g")
+				kernels=$(echo -e "${kernels}" | sed -e "s/\// /g")
 
 				clist="${clist} ${config}"
 				klist="${klist} ${kernels}"
@@ -615,7 +617,7 @@ pass_config_kernel_registries()
 		#   to their respective registries, as appropriate.
 
 		# Handle singleton and umbrella configuration entries separately.
-		if [ $(is_singleton_family "${cname}" "${clist}") == "true" ]; then
+		if [[ $(is_singleton_family "${cname}" "${clist}") == "true" ]]; then
 
 			# Singleton configurations/families.
 			# Note: for singleton families, clist contains one item, which
@@ -625,7 +627,7 @@ pass_config_kernel_registries()
 			# Only consider updating the indirect blacklist (pass 0) or
 			# committing clist and klist to the registries (pass 1) if the
 			# configuration name (cname) is not blacklisted.
-			if [ $(is_in_list "${cname}" "${all_blist}") == "false" ]; then
+			if [[ $(is_in_list "${cname}" "${all_blist}") == "false" ]]; then
 
 				if [ "${passnum}" == "0" ]; then
 					# Even if the cname isn't blacklisted, one of the requisite
@@ -637,7 +639,7 @@ pass_config_kernel_registries()
 					# system architecture.) Thus, we add cname to the indirect
 					# blacklist.
 					for item in ${klist}; do
-						if [ $(is_in_list "${item}" "${config_blist}") == "true" ]; then
+						if [[ $(is_in_list "${item}" "${config_blist}") == "true" ]]; then
 							indirect_blist="${indirect_blist} ${cname}"
 							break
 						fi
@@ -664,7 +666,7 @@ pass_config_kernel_registries()
 
 			# First we check cname, which should generally not be blacklisted
 			# for umbrella families, but we check anyway just to be safe.
-			if [ $(is_in_list "${cname}" "${all_blist}") == "false" ]; then
+			if [[ $(is_in_list "${cname}" "${all_blist}") == "false" ]]; then
 
 				if [ "${passnum}" == "1" ]; then
 
@@ -673,7 +675,7 @@ pass_config_kernel_registries()
 					# omit it from clist and klist.
 					for item in ${clist}; do
 
-						if [ $(is_in_list "${item}" "${all_blist}") == "true" ]; then
+						if [[ $(is_in_list "${item}" "${all_blist}") == "true" ]]; then
 							clist=$(remove_from_list "${item}" "${clist}")
 							klist=$(remove_from_list "${item}" "${klist}")
 						fi
@@ -695,7 +697,7 @@ pass_config_kernel_registries()
 
 	if [ "${passnum}" == "0" ]; then
 		# Assign the final indirect blacklist (with whitespace removed).
-		indirect_blist="$(canonicalize_ws ${indirect_blist})"
+		indirect_blist=$(canonicalize_ws "${indirect_blist}")
 	fi
 }
 
@@ -736,11 +738,11 @@ read_registry_file()
 
 			config=${cr_var##config_registry_}
 
-			clist=$(query_array "config_registry" ${config})
+			clist=$(query_array "config_registry" "${config}")
 
 			# The entries that define singleton families should never need
 			# any substitution.
-			if [ $(is_singleton_family "${config}" "${clist}") == "true" ]; then
+			if [[ $(is_singleton_family "${config}" "${clist}") == "true" ]]; then
 				continue
 			fi
 
@@ -749,7 +751,7 @@ read_registry_file()
 			for mem in ${clist}; do
 
 				#mems_mem="${config_registry[${mem}]}"
-				mems_mem=$(query_array "config_registry" ${mem})
+				mems_mem=$(query_array "config_registry" "${mem}")
 
 				# If mems_mem is empty string, then mem was not found as a key
 				# in the config list associative array. In that case, we continue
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -572,12 +573,13 @@ pass_config_kernel_registries()
...
 		# and the kernel set needed by that configuration are different.
-		if [[ "${list}" == *[/]* ]]; then
+		if [[ ${list} = */* ]]; then
 
 			#echo "Slash found."
 			klist=""
 			clist=""
-			for item in "${list}"; do
+
+			for item in ${list}; do
```
leekillough, line 583: This is a bug -- single quotes makes the loop only execute once, even if `${list}` contains space-separated items.

If the items in `${list}` may need to contain spaces, then declaring `${list}` a Bash array variable and using it such as with `"${list[@]}"` will work.

configure
```diff
@@ -2583,7 +2589,7 @@ main()
 	# -- Command line option/argument parsing ----------------------------------
 
 	found=true
-	while $found = true; do
```
leekillough, line 2586: Bug.

devinamatthews:
@leekillough thanks for going through and fixing all of these! I'll let Field comment on individual changes (@fgvanzee unless you want me to go through it, no problem). I'll test on MacOS since it is stuck on GPLv2 bash.

fgvanzee:
@leekillough Thank you for this exhaustive review. I sincerely appreciate your efforts. I'll go through your list more closely and see if I have any comments.

configure
fgvanzee: Agreed, this was a bug. I think we might have been lucky in that the `config_registry` didn't make use of syntax that would have needed the loop.

fgvanzee:
This looks good, @leekillough. I only made a few trivial whitespace/formatting changes. I'll merge shortly. Thanks again for your contribution.

fgvanzee:
@leekillough Lol, sorry. I got distracted while waiting for CI to finish. I mean it this time! 😅

PR #732, opened 2023-03-25
Omit -fPIC if shared library build is disabled.

fgvanzee:
Details:
- Updated `common.mk` so that when `--disable-shared` option is given to `configure`:
  1. The `-fPIC` compiler flag is omitted from the individual configuration family members' `CPICFLAGS` variables (which are initialized in each subconfig's `make_defs.mk` file); and
  2. The `BUILD_SYMFLAGS` variable, which contains flags needed to control the symbol export behavior, is left blank. 
- The net result of these changes is that flags specific to shared library builds are only used when a shared library is actually being built. Thanks to Nick Knight for reporting this issue.
- `CREDITS` file update.

cc: @nick-knight

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -56,6 +56,7 @@ but many others have contributed code, ideas, and feedback, including
   Tony Kelman              @tkelman
   Lee Killough             @leekillough               (Cray)
   Mike Kistler             @mkistler                  (IBM, Austin Research Laboratory)
+  Nick Knight              @nick-knight               (SiFive)
   Ivan Korostelev          @ivan23kor                 (University of Alberta)
   Kyungmin Lee             @kyungminlee               (Ohio State University)
   Michael Lehn             @michael-lehn
--- a/common.mk
+++ b/common.mk
@@ -731,6 +731,9 @@ $(foreach c, $(CONFIG_LIST_FAM), $(eval $(call append-var-for,CWARNFLAGS,$(c))))
 
 # --- Position-independent code flags (shared libraries only) ---
 
+
+ifeq ($(MK_ENABLE_SHARED),yes)
+
 # Emit position-independent code for dynamic linking.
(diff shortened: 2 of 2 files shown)


nick-knight:
Hi @fgvanzee , I think I may have misunderstood the intention here. 

The documentation says:

> `CPICFLAGS` should be assigned flags to enable position independent code (shared library) flags.

This suggested to me that `CPICFLAGS` are the things I want passed to the compiler in the case of `--enable-shared`, but not in the case of `--disable-shared`. It appears to me --- and I could be mistaken --- that something different is happening:

* if `--enable-shared`, our custom `CPICFLAGS` are being overridden in `common.mk`;
* if `--disable-shared`, our custom `CPICFLAGS` are being passed.

At this point, I'm not sure I fully understand the role of `CPICFLAGS` in `make_defs.mk`.

fgvanzee:
@nick-knight Good catch. I think this was a mistake on our part (or a somewhat sloppy hack). I *think* I agree that the `-fPIC` flag should be applied within the `make_defs.mk` of each subconfig -- at least, I can't see any reason why it should be otherwise. And if that were the case, `common.mk` doesn't need to override or append any flags (although it appears in the case of Windows builds, `-fPIC` needs to be removed).

I can submit a new PR that moves the `-fPIC` flag insertion to the `make_defs.mk` files.

fgvanzee:
@nick-knight I now realize that `common.mk` was doing something slightly less bizarre, which was *appending* the `-fPIC` flag to whatever the subconfigs set within their individual `make_defs.mk` files (although for Windows, the appending behavior was *not* doing what it was supposed to, and was only saved by the fact that the subconfigs' `CPICFLAGS` variables were all blank! 😂 ).

At this point, I don't think `common.mk` has any business appending anything to the `CPICFLAGS` variables, so I'm going to go ahead with the separate PR.

PR #735, opened 2023-03-28
Optionally disable thread-local storage.

fgvanzee:
Details:
- Implemented a new configure option, `--disable-tls`, which allows the user to optionally disable the use of thread-local storage qualifiers on `static` variables in BLIS. This option will rarely be needed, but in some situations may allow BLIS to compile when TLS is unavailable. Like the `--disable-system` option, `--disable-tls` forcibly disables threading (e.g. forcibly uses `--disable-threading`). Thanks to Nick Knight for suggesting this option.
- Modified `frame/include/bli_lang_defs.h` so that `BLIS_THREAD_LOCAL` is defined to nothing when `BLIS_ENABLE_TLS` is not defined.
- Edited `--disable-system` configure status output for clarity.

@nick-knight @hominhquan You may wish to discuss this feature since you both make use of a similar option, `--disable-system`. 🙂

### DIFF
--- a/build/bli_config.h.in
+++ b/build/bli_config.h.in
@@ -53,6 +53,12 @@
 #define BLIS_DISABLE_SYSTEM
 #endif
 
+#if @enable_tls@
+#define BLIS_ENABLE_TLS
+#else
+#define BLIS_DISABLE_TLS
+#endif
+
 #if @enable_openmp@
 #define BLIS_ENABLE_OPENMP
 #if @enable_openmp_as_def@
--- a/configure
+++ b/configure
@@ -3576,11 +3576,7 @@ main()
 		enable_tls_01=1
 	else
 		echo "${script_name}: disabling thread-local storage (TLS) support."
-		echo "${script_name}: WARNING: disabling thread-local storage forcibly disables all threading!"
 		enable_tls_01=0
-
-		# Force threading to be disabled.
-		threading_model='off'
 	fi
 
 	# Check the threading model flag and standardize its value, if needed.
--- a/configure
+++ b/configure
@@ -199,6 +199,16 @@ print_usage()
                  library. When disabled, this option also forces the use
                  of --disable-threading.
 
+   --enable-tls, --disable-tls
+
+                 Enable thread-local storage (TLS) for static variables
+                 in BLIS. The default state is enabled. However, like with
+                 --disable-system, there may be rare situations, such as
+                 when --disable-system is appropriate, where thread-local
+                 storage is unsupported by the compiler. In those cases,
+                 disabling TLS may be a suitable workaround. When disabled,
+                 this option also forces the use of --disable-threading.
+
    --disable-pba-pools, --enable-pba-pools
    --disable-sba-pools, --enable-sba-pools
 
@@ -2512,6 +2522,9 @@ main()
 	# The system flag.
 	enable_system='yes'
 
+	# The thread-local storage flag.
+	enable_tls='yes'
+
 	# The threading flag.
 	threading_model='off'
 
@@ -2690,6 +2703,13 @@ main()
 							;;
 						disable-system)
 							enable_system='no'
+							;;
+
+						enable-tls)
+							enable_tls='yes'
+							;;
+						disable-tls)
+							enable_tls='no'
 							;;
 
 						enable-threading=*)
@@ -3543,8 +3563,21 @@ main()
 		enable_system_01=1
 	else
 		echo "${script_name}: disabling operating system support."
-		echo "${script_name}: WARNING: all threading will be disabled!"
+		echo "${script_name}: WARNING: disabling OS support forcibly disables all threading!"
 		enable_system_01=0
+
+		# Force threading to be disabled.
+		threading_model='off'
+	fi
+
+	# Check if we are building with or without thread-local storage support.
+	if [[ ${enable_tls} = yes ]]; then
+		echo "${script_name}: enabling thread-local storage (TLS) support."
+		enable_tls_01=1
+	else
+		echo "${script_name}: disabling thread-local storage (TLS) support."
+		echo "${script_name}: WARNING: disabling thread-local storage forcibly disables all threading!"
+		enable_tls_01=0
 
 		# Force threading to be disabled.
 		threading_model='off'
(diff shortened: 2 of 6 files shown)


fgvanzee:
> And now I find it a bit strange to say: "We disable multi-threading because we don't have TLS". It's like leaving other low-end embedded architectures behind the door, just because they are not "well-equipped" enough.

@hominhquan Thank you for sharing your honest opinion! I'm happy to report that my choice to disable threading when disabling TLS was not one I was in any way confident of -- I only did it out of (a) caution, and (b) I was copying and pasting code from the part of `configure` that dealt with `--disable-system`. So I'm happy to remove that part if it makes sense to someone with more experience in these matters than me. 🙂

nick-knight (approved):
Thanks @fgvanzee for adding this configure knob.

I don't have any strong opinions (yet) on the broader question of how BLIS uses TLS, since I don't fully understand how it's used.

I think it is reasonable to have this configuration option, since TLS support can be quite involved. For example, [GNU docs](https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html)  say

> It requires significant support from the linker (`ld`), dynamic linker (`ld.so`), and system libraries (`libc.so` and `libpthread.so`), so it is not available everywhere.

hominhquan:
> I only did it out of (a) caution, and (b) I was copying and pasting code from the part of configure that dealt with --disable-system. 

@fgvanzee You are right to care about race-condition, but here I'm curious about write access to `bli_l3_ind_oper_st[]` being already [protected](https://github.com/flame/blis/blob/04090df01175477394d1e73af2e5769751d47cd6/frame/3/bli_l3_ind.c#L56) (I would protect the read-access as well, or use lock-free atomic), but the array is still put in TLS, why ? because it is not needed to protect a TLS variable, unless that TLS variable is shared to another thread ? (nooooo ... )

> @hominhquan Is there such as system that supports multithreading but not TLS? It seems like the latter is a prerequisite for most implementations of the former? 

TLS is not a must-have feature to run multithreading code. The proof is BLIS has run well in either OpenMP or Pthread on C6x, or MPPA and some others before. On another hand, a sequential single-thread code can totally put a variable in TLS, and that code will compile and work on most CPU today. Most, but not all.

Supporting TLS on new architectures/OS/compilers is often a hazardous and hackish task, which implies many architecture-specific handling, e.g. https://github.com/zephyrproject-rtos/zephyr/issues/22185

> I certainly have no objection to disabling it if possible, but there is some useful functionality which it enables. Even if the library were architected to be free of global state, a globally-stateful interface would be the most natural for a lot of people, especially when they have to access BLIS through the BLAS interface for legacy reasons.

Totally agree and I have no problem (for legacy and historical reasons) with these global structures between the BLAS and BLIS API.

My rationale is: Today we are writing code that will run on who-knows-how-many different architectures in the next years/decades. I don't know if they will be extremely sophisticated, or contrastingly very simple and poorly-equipped for power reason and/or application-domain. Using TLS is one way to make our code less portable.

fgvanzee:
> @fgvanzee You are right to care about race-condition, but here I'm curious about write access to bli_l3_ind_oper_st[] being already [protected](https://github.com/flame/blis/blob/04090df01175477394d1e73af2e5769751d47cd6/frame/3/bli_l3_ind.c#L56) (I would protect the read-access as well, or use lock-free atomic), but the array is still put in TLS, why ? because it is not needed to protect a TLS variable, unless that TLS variable is shared to another thread ? (nooooo ... )

IIRC, our collaborators at AMD once pointed out that TLS was desired to enforce some determinism on `static` variables such as `bli_l3_ind_oper_st[]` (the other example being error checking level I think?) because of shenanigans that could take place in application-level threading (rather than BLIS-level threading). It was certainly not a mainstream use case, but rather a situation where they wanted to prevent accidental misuse.

devinamatthews:
@hominhquan @leekillough thanks for pointing out that TLS is more difficult to support than multithreading. If TLS were to be disabled but not multithreading, as a user would you consider it acceptable for thinks like number of threads to spawn, 1m enable/disabled, error checking level, etc. to be global (and possibly subject to race conditions)? Is it instead possible to fake the TLS behavior using the ptheads functions mentions by @leekillough ?

hominhquan:
> TLS was desired to enforce some determinism on static variables such as bli_l3_ind_oper_st[] (the other example being error checking level I think?) because of shenanigans that could take place in application-level threading (rather than BLIS-level threading)

OK, thanks for explanation @fgvanzee.

> @hominhquan @leekillough thanks for pointing out that TLS is more difficult to support than multithreading. If TLS were to be disabled but not multithreading, as a user would you consider it acceptable for thinks like number of threads to spawn, 1m enable/disabled, error checking level, etc. to be global (and possibly subject to race conditions)? 

@devinamatthews @fgvanzee Between (1) disable both TLS and multithreading for safety and (2) disable TLS but not multithreading at the risk of data-race. I would regrettingly choose the solution (1). In any case, functionality comes first. Debugging multithreading data-race is horribly a headache.

I still hope these runtime logic modifications can be reimplemented differently without recourse to TLS. At least it will be the first thing I do when if happens to me to rebase the latest BLIS version.

> Is it instead possible to fake the TLS behavior using the ptheads functions mentions by @leekillough ?

IMHO, supporting explicit `thread_local` using `pthread` primitives mentioned by @leekillough  is equivalent to using malloc/free or internal memory pool mechanism in BLIS, which is also probably what I would use to get rid of TLS.

fgvanzee:
> @devinamatthews @fgvanzee Between (1) disable both TLS and multithreading for safety and (2) disable TLS but not multithreading at the risk of data-race. I would regrettingly choose the solution (1). In any case, functionality comes first. Debugging multithreading data-race is horribly a headache.

In the abstract, I would agree. But I could also make the case that disabling TLS is a very "I know what I'm doing" kind of thing. The only people who would be doing so should be aware of the possible consequences for multithreading, especially if I include a warning in the `configure --help` text, which I am happy to.

> I still hope these runtime logic modifications can be reimplemented differently without recourse to TLS. At least it will be the first thing I do when if happens to me to rebase the latest BLIS version.

I'm always open to making BLIS more stateless, where possible. Maybe such hypothetical changes don't completely eliminate the applicability of TLS, but they may meaningfully reduce it.

hominhquan:
> In the abstract, I would agree. But I could also make the case that disabling TLS is a very "I know what I'm doing" kind of thing. The only people who would be doing so should be aware of the possible consequences for multithreading, especially if I include a warning in the configure --help text, which I am happy to do.

So OK to keep multithreading and add a warning message about possible data-race, and accordingly recommend the developer to try `--disable-threading` if (s)he encounters any correctness or deadlock issue. Such message can be added in both `configure` and the test runtime inside the macro `#ifdef BLIS_DISABLE_TLS` (at the beginning ?)

fgvanzee:
@hominhquan I opted not to insert a compile-time warning since the warning would be emitted for *every since* file compilation (since everything goes through the one header, `blis.h`).

fgvanzee:
Are we all okay now with this compromise? (Leave threading untouched when TLS is disabled, but warn of consequences in `./configure --help`, in `configure` output, and in the testsuite informational header output.)

devinamatthews:
I guess.

hominhquan:
LGTM.

PR #736, opened 2023-03-29
Add output.testsuite to .gitignore since it isn't matched by output.testsuite.*

leekillough:
Add `output.testsuite` to `.gitignore` since it isn't matched by `output.testsuite.*`. I accidentally added `output.testsuite` in a commit with `git add .` after a test run.


leekillough:
ping @fgvanzee

fgvanzee:
Thanks @leekillough!

PR #738, opened 2023-03-29
Move -fPIC insertion to subconfigs' make_defs.mk.

fgvanzee:
Details:
- Previously, `common.mk` was appending `-fPIC` to the `CPICFLAGS` variables set within the various subconfigurations' `make_defs.mk` files. This seemed somewhat unintuitive, and so now the `-fPIC` flag is assigned to the various subconfigs' `CPICFLAGS` variables in the respective `make_defs.mk` files.
- This also commit changes the logic in `common.mk` so that instead of appending, the variable is overwritten, but now *only* in the case of Windows (since apparently `-fPIC` needs to be omitted there). Thanks to Nick Knight for catching and reporting this weirdness.

@nick-knight

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -731,9 +731,6 @@ $(foreach c, $(CONFIG_LIST_FAM), $(eval $(call append-var-for,CWARNFLAGS,$(c))))
 
 # --- Position-independent code flags (shared libraries only) ---
 
-
-ifeq ($(MK_ENABLE_SHARED),yes)
-
 # Note: Avoid -fPIC flags for Windows builds since all code is position-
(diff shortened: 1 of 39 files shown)


nick-knight (approved):
Thanks-- this makes much more sense to me than it did before. But, seeing the changes you've made to essentially everyone's `make_defs`, it seems like I'm the only one who was confused by this, and I feel a bit guilty.

fgvanzee:
@nick-knight No, you should feel the opposite of guilty. Nobody else caught it, going back at least five years!

PR #743, opened 2023-05-04
Fix enablement of 1m for syrk and friends in the testsuite

devinamatthews:
Because (sy|he)r2?k are implemented in terms of gemmt, the testsuite doesn't actually turn on 1m for these operations by enabling induced methods for the corresponding operation ID. This commit provides a somewhat hackish workaround.

Alternatively, it seems like it wouldn't be harmful to just enable the induced method for all operations while testing.


fgvanzee:
Dang, that *is* hacky... Let me think about this a bit longer.

> Alternatively, it seems like it wouldn't be harmful to just enable the induced method for all operations while testing.

But wouldn't that mean that the native implementations of level-3 operations don't get tested?

PR #744, opened 2023-05-04
Simplify the GENTFUNC/GENTPROT macros using variadic macros.

devinamatthews:
Variadic macros are a standard C99 feature and widely supported. I took care not to use variadic macros where 0 variadic arguments are expected as that is a non-standard extension.


fgvanzee:
Looks good to me. I'll draft a commit log entry and squash shortly.

PR #747, opened 2023-05-06
add missing include for windows

### LINKED ISSUE #746: BUG: remove use of long-deprecated implicit function declarations (compat with newer clang)
ISO C23 finally removed K&R-style, implicit function declarations, and clang 16 now got [stricter](https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#id85) about enforcing the long-standing deprecation of these (introduced already as deprecated in the first C standard 🤯)

This causes compilation of blis to fail as follows:

```
blastest/f2c/open.c:180:7: fatal error: call to undeclared function '_access'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                if (access(buf,0))
                    ^
blastest/f2c/open.c:31:16: note: expanded from macro 'access'
#define access _access
```

AFAIU, GCC is planning to enable a similar warning by default starting in GCC 14.

h-vetinari:
Fixes #746, confirmed in https://github.com/conda-forge/blis-feedstock/pull/28

As specified by the MSVC [docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/access-waccess?view=msvc-170#requirements) for using `_access`.

### DIFF
--- a/blastest/f2c/open.c
+++ b/blastest/f2c/open.c
@@ -28,6 +28,7 @@ use or performance of this software.
 #include <unistd.h>
 #endif
 #ifdef _MSC_VER
+#include <io.h>
 #define access _access
 #endif
 #include "f2c.h"


fgvanzee:
@h-vetinari Thanks for this PR. Looks fine to me (and to AppVeyor), so I'll merge.

PR #755, opened 2023-07-19
Exclude -lrt on Android platforms with Bionic libraries

### LINKED ISSUE #752: How to cross-compile libraries for use on Android?
when complie blis using Android NDK 23, error is `ld : error: unable to find library -lrt`
```
./configure --enable-threading=pthreads CC=$NDK_PATH/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang CXX=$NDK_PATH/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang++  AR=$NDK_PATH/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-ar RANLIB=$NDK_PATH/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-ranlib LIBPTHREAD=$NDK_PATH/toolchains/llvm/prebuilt/linux-x86_64/lib64/libc++.so arm64

make -j32
Dynamically linking lib/arm64/libblis.so
Archiving lib/arm64/libblis.a
ld: error: unable to find library -lrt
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
Makefile:790: recipe for target 'lib/arm64/libblis.so' failed
make: *** [lib/arm64/libblis.so] Error 1
```

libpthread and librt libraries were included in libc
https://developer.android.com/ndk/guides/stable_apis#c_library

How to quickly integrate blis on Android？

leekillough:
Fixes #752, by excluding `-lrt` when Android Bionic system libraries are detected with the `__BIONIC__` preprocessor macro.

This is not a BLIS `configure` parameter per se; it is just a modification to the `Makefile` build system to exclude adding `-lrt` during Linux builds if an Android system is detected.

This needs to be tested thoroughly. I do not have an Android system to test it with. It is fairly conservative, in that it should not change the behavior of BLIS builds unless `__BIONIC__` is defined, and as long as `$(CC)` is executable during the parsing of `common.mk`. 

Cc: @leo4678


fgvanzee:
Thanks for this PR, @leekillough. It looks good to me.

@leo4678 Once you confirm that this addresses your situation, I'll merge.

leekillough:
I wonder where the `bionic.h` header file should be located; I don't want to clutter the top-level BLIS directory.

Alternatively, I even thought about making the GNU Makefile automatically create the `bionic.h` file on-the-fly. I think there are some `bash` "heredoc" features which make this easier, but it's hard to put them inside of a Makefile, and it might assume that `bash` or another shell is the underlying shell which `make` uses.

So maybe a `bash` script which is invoked by the Makefile would be simpler. Again, I wouldn't want it to clutter the top-level BLIS directory.

fgvanzee:
I appreciate your thoughtfulness on this, @leekillough. I also would prefer to not add the file to the top-level directory if at all possible.

devinamatthews:
Why not put this in `configure`?

leekillough:
> Why not put this in `configure`?

It is not significant enough of a change to warrant a new `configure` option, and if it is added to `configure`, it adds more testing requirements.

It's more like: "For Linux builds, regardless of target architecture, we normally link with `-lrt`, but on Android, this library's features are implicit in Bionic, and `-lrt` cannot be explicitly linked with. So on Android Linux builds, we make an exception for `-lrt`, but keep everything else the same, including `configure`".

I'd like a suggestion of where best to place the `bionic.h` file so that it doesn't clutter the main directory.

fgvanzee:
> I'd like a suggestion of where best to place the bionic.h file so that it doesn't clutter the main directory.

How about creating a new directory within `build/detect` and placing it there?

PR #759, opened 2023-07-26
small fixes to support hpx in the testsuite

### LINKED ISSUE #758: hpx runtime support breaks in testsuite
the hpx runtime support is currently broken in the blis testsuite. the testsuite segmentation faults due to posix thread calls conflicting with the hpx runtime system. the solution requires removing posix threads in testsuite for the hpx backend.

ct-clmsn:
This PR resolves #758. Places compile time guards around the posix portions of the test suite.


leekillough (approved):
LGTM

PR #767, opened 2023-08-15
Revamped bli_init() to use TLS where feasible.

fgvanzee:
Details:
- Revamped `bli_init_apis()` and `bli_finalize_apis()` to use separate `bli_pthread_switch_t` objects for each of the five sub-API init functions, with the objects for the 'ind' and 'rntm' sub-APIs being declared with `BLIS_THREAD_LOCAL`. This allows some APIs to be treated as thread-local and the rest as thread-shared. Thanks to Edward Smyth for requesting application thread-specific `rntm_t` structs, which inspired these change.
- Combined `bli_thread_init_from_env()` and `bli_pack_init_from_env()` into a new function, `bli_rntm_init_rntm_from_env()`, and placed the combined code in `bli_rntm.c` inside of a new `bli_rntm_init()` function. Then removed the (now empty) `bli_pack_init()` and `_finalize()` function defs.
- Deprecated `bli_rntm_init()` for the purposes of initializing a `rntm_t` (temporarily preserving it as cpp-undefined code as `bli_rntm_clear()`) so that the function name could be used for the aforementioned `bli_rntm_init()` function.
- Updated `libblis_test_pobj_create()` in `test_libblis.c` to use a static `rntm_t` initializer instead of the deprecated `bli_rntm_init()` function-based option.
- Minor updates to `docs/Multithreading.md`, including removal of `bli_rntm_init()` in the example of how to initialize `rntm_t` structs.
- Changed the return value of `bli_gks_init()`, `bli_ind_init()`, `bli_memsys_init()`, `bli_thread_init()`, and `bli_rntm_init()` (and their `_finalize()` counterparts) from `void` to `int` so that those functions match the function type expected by `bli_pthread_switch_on()`/`_off()`. Those init/finalize functions now return 0 to indicate success, which is needed so that the switch actually changes state from off to on and vice versa.
- Defined `bli_thread_reset()`, which copies the contents of the `global_rntm_at_init` struct into the `global_rntm` struct (for the current application thread).
- Guard calls to `bli_pthread_mutex_lock()`/`_unlock()` in
  - `bli_pack_set_pack_a() and _pack_b()`
  - `bli_rntm_init_from_global()`
  - `bli_thread_set_ways()`
  - `bli_thread_set_num_threads()`
  - `bli_thread_set_thread_impl()`
  - `bli_thread_reset()`
  - `bli_l3_ind_oper_set_enable()` with `#ifdef BLIS_DISABLE_TLS` (since TLS precludes the possibility of race conditions).
- In `frame/base/bli_rntm.c`, declare `global_rntm`, `global_rntm_at_init`, and `global_rntm_mutex` as `BLIS_THREAD_LOCAL` so that separate application threads can change the number of ways of BLIS parallelism independently from one another.
- Access `global_rntm` only via a new private (not exported) function, `bli_global_rntm()`. Defined a similar function for a `rntm_t` new to this commit, `global_rntm_at_init`, which preserves the state of the global rntm at initialization-time.
- In `frame/3/bli_l3_ind.c`, added a guard to the declaration of the static variable `oper_st_mutex` with `#ifdef BLIS_DISABLE_TLS` so that the mutex is omitted altogether when TLS is enabled (which prevents the compiler from warning about an unused variable).
- Removed redundant code from `bli_thread.c`:
   ```c
    #ifdef BLIS_ENABLE_HPX
    #include "bli_thread_hpx.h"
    #endif
   ```
  since this code is already present in `bli_thread.h`.
- Comment updates.

cc: @edwsmyth

### DIFF
--- a/docs/Multithreading.md
+++ b/docs/Multithreading.md
@@ -205,6 +205,8 @@ If you still wish to set the parallelization scheme globally, but you want to do
 
 **Note**: Regardless of which way ([automatic](Multithreading.md#globally-at-runtime-the-automatic-way) or [manual](Multithreading.md#globally-at-runtime-the-manual-way)) the global runtime API is used to specify multithreading, that specification will affect operation of BLIS through **both** the BLAS compatibility layer as well as the native ([typed](docs/BLISTypedAPI.md) and [object](docs/BLISObjectAPI.md)) APIs that are unique to BLIS.
 
+If BLIS is being used by two or more application-level threads, each of those application threads will track their own global state for the purpose of specifying parallelism. We felt this makes sense because each application thread may wish to specify a different parallelization scheme without affecting the scheme for he other application thread(s).
+
 ### Globally at runtime: the automatic way
 
 If you simply want to specify an overall number of threads and let BLIS choose a thread factorization automatically, use the following function:
@@ -280,10 +282,6 @@ You **must** initialize the `rntm_t`. This can be done in either of two ways.
 If you want to initialize it as part of the declaration, you may do so via the default `BLIS_RNTM_INITIALIZER` macro:
 ```c
 rntm_t rntm = BLIS_RNTM_INITIALIZER;
-```
-Alternatively, you can perform the same initialization by passing the address of the `rntm_t` to an initialization function:
-```c
-bli_rntm_init( &rntm );
 ```
 As of this writing, BLIS treats a default-initialized `rntm_t` as a request for single-threaded execution.
 
--- a/docs/Multithreading.md
+++ b/docs/Multithreading.md
@@ -205,7 +205,7 @@ If you still wish to set the parallelization scheme globally, but you want to do
 
 **Note**: Regardless of which way ([automatic](Multithreading.md#globally-at-runtime-the-automatic-way) or [manual](Multithreading.md#globally-at-runtime-the-manual-way)) the global runtime API is used to specify multithreading, that specification will affect operation of BLIS through **both** the BLAS compatibility layer as well as the native ([typed](docs/BLISTypedAPI.md) and [object](docs/BLISObjectAPI.md)) APIs that are unique to BLIS.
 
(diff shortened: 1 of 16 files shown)


docs/Multithreading.md
```diff
@@ -205,6 +205,8 @@ If you still wish to set the parallelization scheme globally, but you want to do
 
 **Note**: Regardless of which way ([automatic](Multithreading.md#globally-at-runtime-the-automatic-way) or [manual](Multithreading.md#globally-at-runtime-the-manual-way)) the global runtime API is used to specify multithreading, that specification will affect operation of BLIS through **both** the BLAS compatibility layer as well as the native ([typed](docs/BLISTypedAPI.md) and [object](docs/BLISObjectAPI.md)) APIs that are unique to BLIS.
 
+If BLIS is being used by two or more application-level threads, each of those application threads will track their own global state for the purpose of specifying parallelism. We felt this makes sense because each application thread may wish to specify a different parallelization scheme without affecting the scheme for he other application thread(s).
```
hominhquan, line 208: typo "for he other" ?

frame/3/bli_l3_ind.c
```diff
@@ -36,7 +36,7 @@
 #include "blis.h"
 
 // This array tracks whether a particular operation is implemented for each of
-// the induced methods.
+// the induced methods. This array is meant to be read-only.
```
hominhquan, line 39: If read-only, I would add `const` as well.

docs/Multithreading.md
fgvanzee: Yes, thank you!

fgvanzee:
@edwsmyth and @kiran-amd have signed off on this. Anyone else have any comments before I merge?

hominhquan:
> @edwsmyth and @kiran-amd have signed off on this. Anyone else have any comments before I merge?

No comment :)

fgvanzee:
Thank you everyone for your feedback on this! I'll merge later today or tomorrow.

PR #774, opened 2023-09-20
Update Multithreading.md

jmather-sesi:
Replaced 404'd link with an archive from The Wayback Machine.


fgvanzee:
Thank you for catching this, @jmather-sesi!

PR #782, opened 2023-10-11
Fixed bug in sup threshold registration.

### LINKED ISSUE #781: Default BLIS_[MNK]T values never actually set
In `bli_cntx_init_ref.c`, the default values for the gemmsup thresholds (`BLIS_[MNK]T` blocksizes) are set to zero, so that no operation ever matches the criteria for gemmsup. HOWEVER, these are set via `bli_cntx_set_blkszs` which calls `bli_cntx_set_blksz` which calls `bli_blksz_copy_if_pos`. Thus, the zero values are simply discarded and those blocksizes remain uninitialized. The upshot of this is that the reference gemmsup handler is called for gemm problems essentially at random (and as it turns out, very rarely the reference gemmsup implementation encounters a divide-by-zero error). Because `bli_gemmt?sup` checks if m, n, and k are *less than* the thresholds, setting the default values to 1 should be safe and effective.

fgvanzee:
Details:
- Fixed a bug that resulted in BLIS non-deterministically calling the `gemmsup` handler, irrespective of the thresholds that are registered via `bli_cntx_set_blkszs()`.
- Deep dive: In `bli_cntx_init_ref.c`, the default values for the `gemmsup` thresholds (`BLIS_[MNK]T` blocksizes) were being set to zero so that no operation ever matched the criteria for `gemmsup` (unless specific sup thresholds are registered). HOWEVER, these thresholds are set via `bli_cntx_set_blkszs()` which calls `bli_blksz_copy_if_pos()`, which was only coping the thresholds into the gks' `cntx_t` if the values were strictly positive. Thus, the zero values passed into `bli_cntx_set_blkszs()` were being ignored and those threshold slots within the gks were left uninitialized. The upshot of this is that the reference `gemmsup` handler was being called for `gemm` problems essentially at random (and as it turns out, very rarely the reference `gemmsup` implementation would encounter a divide-by-zero error).
- The problem was fixed by changing `bli_blksz_copy_if_pos()` so that it copies values that are non-negative (values >= 0 instead of > 0). The function was also renamed to `bli_blksz_copy_if_nonneg()`.
- Fixes #781. Thanks to Devin Matthews for identifying, diagnosing, and proposing a fix for this issue.

### DIFF
--- a/config/bgq/bli_cntx_init_bgq.c
+++ b/config/bgq/bli_cntx_init_bgq.c
@@ -69,11 +69,11 @@ void bli_cntx_init_bgq( cntx_t* cntx )
 
 	// Initialize level-3 blocksize objects with architecture-specific values.
 	//                                           s      d      c      z
-	bli_blksz_init_easy( &blkszs[ BLIS_MR ],     0,     8,     0,     4 );
-	bli_blksz_init_easy( &blkszs[ BLIS_NR ],     0,     8,     0,     4 );
-	bli_blksz_init_easy( &blkszs[ BLIS_MC ],     0,  1024,     0,   768 );
-	bli_blksz_init_easy( &blkszs[ BLIS_KC ],     0,  2048,     0,  1536 );
-	bli_blksz_init_easy( &blkszs[ BLIS_NC ],     0, 10240,     0, 10240 );
+	bli_blksz_init_easy( &blkszs[ BLIS_MR ],    -1,     8,    -1,     4 );
+	bli_blksz_init_easy( &blkszs[ BLIS_NR ],    -1,     8,    -1,     4 );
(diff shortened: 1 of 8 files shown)


devinamatthews:
There was an FPE in the Penryn SDE test.

fgvanzee:
Yeah I saw. I'm able to reproduce it on my x86_64 system with `./configure penryn; make; make check`.

fgvanzee:
Ah, I see. In the `bli_cntx_init*()` functions, we use 0 and -1 interchangeably to indicate "don't register a value for this one". But when we use 0 (as in the case of `penryn`) that 0 is now copied over under this PR's changes, whereas before it was ignored.

So we could:
1. Change all of the 0's to -1's.
2. Backtrack and use 1 instead of 0 like you originally proposed, and deal with any weird "sup now handles edge cases🤞🏼" behavior later.

devinamatthews:
I vote #1

PR #784, opened 2023-10-30
Allow users to defines [sd]complex using std::complex

devinamatthews:
In C++ applications, it makes a lot of sense to interface to BLIS using C++'s standard complex number library, which uses a template class std::complex. Obviously BLIS doesn't know anything about this and defaults to a custom struct to represent complex numbers. This PR updates the bli_[cz]{real,imag} functions to accept std::complex numbers when a C++ compiler is being used. Note that this has no effect on the compilation of the BLIS library (or testsuite), and only comes into play when including `blis.h` into a C++ project and forcing the use of std::complex for scomplex and dcomplex.

A bug in the definition of some scalar level-0 macros was also fixed, since bli_creal/bli_cimag and bli_zreal/bli_zimag are no longer interchangeable.


Aaron-Hutchinson:
This compatibility layer is greatly appreciated! Thanks for adding it.

devinamatthews:
@Aaron-Hutchinson BTW the mechanism to override BLIS's typedefs of [sd]complex is currently undocumented (actually, it's a holdover from libflame integration). As part of this PR I guess I should add some documentation and maybe introduce a more user-friendly override macro.

fgvanzee:
@devinamatthews Remind me: this is not yet ready for me to review yet, right? I think you said you were going to work on it further...? 🤔

devinamatthews:
Yes I need to rework some stuff.

devinamatthews:
@fgvanzee now this is ready to review and merge.

fgvanzee:
> @fgvanzee now this is ready to review and merge.

Looks good to me.

Can you remind me, where would `BLIS_ENABLE_STD_COMPLEX` get defined? Is this the kind of thing that would ideally be a `configure` option, but for now would have to be `#defined` in the target configuration's `bli_family_*.h` headers?

devinamatthews:
This would be in user code e.g.
```c++
#define BLIS_ENABLE_STD_COMPLEX
#include <blis.h>
//call blis functions using std::complex<double>
```

Issue #786, opened 2023-11-06
Header path for default source build and Debian should match


jedbrown:
A default source build installs `$incdir/blis/blis.h`, suggesting that a user should `#include <blis/blis.h>`.

Debian installs `/usr/include/x86_64-linux-gnu/blis.h`, which works with `#include <blis.h>`. (This path is actually a symlink through the `/etc/alternatives` system, but that aspect does not concern us here.)

Upstream should decide (and convey to users) which schema is intended. @devinamatthews and @fgvanzee [have indicated in Discord](https://discord.com/channels/879093420801527858/879093420801527861/1170117964767445034) that `#include <blis.h>` is preferred, meaning that the default `make install` should be changed. Although symlinks `blis.h -> blis/blis.h` are problematic on Windows, the outer header could be a simple stub with `#include "blis/blis.h"`.

jedbrown:
With 1.0 coming up, this should be resolved and documented prior to release.

fgvanzee:
Thanks @jedbrown. Devin and I discussed this and we both agree that our preferred solution is to install a supplemental one-line `blis.h` header to the `INCDIR` install prefix (which is `PREFIX/include` by default, where `PREFIX` is the install prefix) that contains:
```c
#include <blis/blis.h>
```
FYI: @cdluminate

fgvanzee:
FYI, I'm working on a PR for this issue.

cdluminate:
I will change the alternatives symlink installation path to synchronize with you.

fgvanzee:
@jedbrown I think this issue is resolved now. Let me know if there's anything left/else to discuss. Thanks for your feedback!

fgvanzee:
@jedbrown Closing this issue for now. Please let us know if you have any further concerns.

PR #787, opened 2023-11-11
Install helper headers to INCDIR prefix.

fgvanzee:
Details:
- Install one-line headers to `INCDIR` whose entire purpose is to `#include` the actual headers within the local `blis` header directory so that applications can `#include "blis.h"` instead of `#include "blis/blis.h"` (and/or `"cblas.h"` instead of `"blis/cblas.h"` if CBLAS is enabled) when headers are installed to global paths. (Note that `INCDIR` is the installation prefix for headers as specified by `--includedir=INCDIR`, which defaults to `PREFIX/include` if not specified.) Not sure how this problem went unreported for so long, since presumably any user trying to `#include "blis.h"` from a global installation would have encountered a compiler error.
- The one-line `blis.h` and `cblas.h` headers now reside in the `build` directory, ready to install as is.
- Thanks to to Jed Brown for reporting this via Issue #786, and for Devin Matthews and Mo Zhou for their engagement.
- Harmonized the rule in the top-level Makefile for installing `blis.pc` into `SHAREDIR/pkgconfig` with conventions for others vis-a-vis verbosity/non-verbosity.

cc: @jedbrown @cdluminate

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,9 +284,12 @@ endif
 # Define a list of headers to install. The default is to only install blis.h.
 HEADERS_TO_INSTALL      := $(BLIS_H_FLAT)
 
-# If CBLAS is enabled, we also install cblas.h so the user does not need to
-# change their source code to #include "blis.h" in order to access the CBLAS
-# function prototypes and enums.
+# If CBLAS is enabled, we also install cblas.h. This allows the user to continue
+# using #include "cblas.h" in their application, if they wish. (NOTE: Even if we
+# didn't install cblas.h, the user could *still* access CBLAS definitions and
+# function prototypes, but they would have to update their source code to use
+# #include "blis.h" instead of #include "cblas.h" since the latter header file
+# would not exist.)
 ifeq ($(MK_ENABLE_CBLAS),yes)
 HEADERS_TO_INSTALL += $(CBLAS_H_FLAT)
 endif
--- a/Makefile
+++ b/Makefile
@@ -1067,7 +1067,7 @@ install-helper-headers: check-env $(HELP_HEADERS_INSTALLED)
 
 # A rule to install a helper header file.
 define make-helper-header-rule
-$(INSTALL_INCDIR)/$(notdir $(1)): $(BUILD_DIR)/$(notdir $(1)) $(CONFIG_MK_FILE)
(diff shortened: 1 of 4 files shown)


devinamatthews:
Helper include files should use <> not "".

Makefile
```diff
@@ -282,7 +282,7 @@ endif
 #
 
 # Define a list of headers to install. The default is to only install blis.h.
-HEADERS_TO_INSTALL := $(BLIS_H_FLAT)
+HEADERS_TO_INSTALL      := $(BLIS_H_FLAT)
 
 # If CBLAS is enabled, we also install cblas.h so the user does not need to
 # change their source code to #include "blis.h" in order to access the CBLAS
```
jedbrown, line 288:
```suggestion
# change their source code to #include <blis.h> in order to access the CBLAS
```
But this comment seems out of place.

jedbrown (commented):
There are many places throughout the repository that `#include "blis.h"` (especially user-facing docs) that should be `#include <blis.h>` because you don't actually want to search a relative path.

Makefile
fgvanzee: Yes, this was likely a copy-paste bug. Thanks Jed, I'll fix it.

fgvanzee:
> Helper include files should use <> not "".

Thanks. I'll change the contents of the helper headers to use `< >`.

Makefile
fgvanzee: I just looked into this, and the mention of `blis.h` is actually *not* a typo. What it's trying to say is the following:

If the user already has an application that has `#include "cblas.h"` and wants to use BLIS as a CBLAS implementation, then we have to install the `cblas.h` header in order to provide the best portability experience. If we didn't install the `cblas.h` header, the user could *still* use BLIS, but they'd have to update their application source code to use `#include "blis.h"` instead of `#include "cblas.h"`.

I'll try to update the comment to clarify.

fgvanzee:
Ugh, BSD `install` doesn't recognize `-T`. 💀

fgvanzee:
Actually, I may not need `-T`... 🤞🏼

fgvanzee:
@jedbrown @cdluminate If this looks good to you, I'll squash-merge, maybe later today.

PR #789, opened 2023-12-05
Include `bli_config.h` before `bli_system.h` in `cblas.h`.

fgvanzee:
Details:
- Previously, in `cblas.h`, `bli_config.h` was being `#included` *after* `bli_system.h`, which meant that the `BLIS_ENABLE_SYSTEM` macro was never defined in time for proper OS detection. This bug only affected `cblas.h` -- `blis.h` had been correctly including `bli_config.h` before `bli_system.h` since fb93d24. Thanks to Edward Smyth for reporting this bug and suggesting the fix.

cc: @edwsmyth

### DIFF
--- a/frame/compat/cblas/src/cblas.h
+++ b/frame/compat/cblas/src/cblas.h
@@ -3,11 +3,16 @@
 #define CBLAS_H
 #include <stddef.h>
 
-// We need to #include "bli_type_defs.h" in order to pull in the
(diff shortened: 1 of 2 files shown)


hominhquan:
Oh, I somehow missed to grep for all `#include "bli_config.h"` occurrence when I submitted the patch. My bad and thanks @fgvanzee for fixing this.

Issue #793, opened 2024-02-11
A more complete list of ARM cpu implementations


j-bm:
The code currently does not identify all recent ARM 64-bit cpu implementations (and a few 32-bit ones too). 

The code also uses hardcoded constants like0xc09 instead of symbols like ARM_CPU_PART_CORTEX_A9.

See frame/base/bli_cpuid.c

A PR to follow.

PR #794, opened 2024-02-11
Add cpu part codes for various manufacturers and use in the code

j-bm:
This addresses issue #793 by collecting cpu part numbers from various sources and adding them or uncommenting them in the code.

A few hardcoded constants are then replaced.

Most of this is for readability.  A copy-paste error is also fixed in one routine, correcting argument names.

Note that one functional change is to setup for all Apple Silicon implementations, not just one of them.


devinamatthews:
Thanks this is a big improvement.

PR #802, opened 2024-03-25
Update BLIS_*_INITIALIZER macros for C++ compatibility.

devinamatthews:
- Remove designated initializer syntax. This isn't officially supported until C++20.
- Put initializers in the order in which they are defined in the struct. Even with standard or extension support for designated initializers, initializing non-static members out-of-order is an error in C++.
- Remove the conditional code which uses `-1` as the default value of the `pack_buf` member of `mem_t` in C, but `BLIS_BUFFER_FOR_GEN_USE` in C++. Simply use the latter as a common-sense default.

### DIFF
--- a/frame/base/bli_mem.h
+++ b/frame/base/bli_mem.h
@@ -136,38 +136,19 @@ BLIS_INLINE void bli_mem_set_size( siz_t size, mem_t* mem )
 // removed from the mem_t type definition. An alternative to the initializer is
 // calling bli_mem_clear() at runtime.
 
-#ifdef __cplusplus
 #define BLIS_MEM_INITIALIZER \
         { \
-          .pblk        = BLIS_PBLK_INITIALIZER, \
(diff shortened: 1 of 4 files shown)


devinamatthews:
@fgvanzee ping

fgvanzee:
This is probably fine, but *boy* does it give me the opposite of a warm-and-fuzzy by taking away the field names from their respective initialization expressions. 😬

leekillough:
As a language lawyer, I'm all for the removal of designated initializers, which is one of the largest differences between C and C++ (the other being anonymous `struct`s).

devinamatthews:
Voila.

fgvanzee:
Mmm, much better. Not ideal (since it's still just a readability aid), but I can live with this. Thanks.

fgvanzee (approved):

fgvanzee:
Anything else before I merge?

devinamatthews:
Go for it.

PR #819, opened 2024-07-10
Flatten `cblas.h` immediately after `blis.h`.

fgvanzee:
Details:
- Previously, if the user enabled CBLAS via `configure --enable-cblas` and then ran `make`, the flattened `blis.h` header file would be created immediately, but the flattened `cblas.h` header file would not be created until `make install` was run. This was happening because nothing in the BLIS build process (except installation) depended on the flattened `cblas.h` (whereas *everything* depends on the flattened `blis.h`, and therefore it was being created first). This behavior can be confusing to application developers who could reasonably expect that the flattened `cblas.h` header would be available (to inspect or use) prior to running 'make install'.
- This commit fixes the aforementioned issue by (1) adding `cblas.h` (if CBLAS is enabled) as a dependency to all of the build rules for core framework object files, and (2) making the flattened `blis.h` a prerequisite for flattening `cblas.h`. The upshot is that (1) ensures that the flattened `cblas.h` is created around the the same time that the flattened `blis.h` is created, and (2) ensures that the two headers are flattened sequentially (first `blis.h` and then `cblas.h`) even when using `make -j[n]`, which ensures that the output of the two processes do not co-mingle.
- Thanks to Jeff Diamond for reporting this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,16 +281,17 @@ endif
 # --- Monolithic header definitions --------------------------------------------
 #
 
-# Define a list of headers to install. The default is to only install blis.h.
+# Define lists of headers to create/install. The default is to only
+# create/install blis.h.
+HEADERS_TO_BUILD        := $(BLIS_H_FLAT)
 HEADERS_TO_INSTALL      := $(BLIS_H_FLAT)
 
-# If CBLAS is enabled, we also install cblas.h. This allows the user to continue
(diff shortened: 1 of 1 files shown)


fgvanzee:
@devinamatthews Any comments or concerns with this? If not, I'll merge.

devinamatthews:
Looks good.

PR #822, opened 2024-08-01
Use intrinsics for all sifive_x280 kernels

myeh01:
To address a bug encountered in #805, this PR converts all kernels in the sifive_x280 configuration written in inline assembly to intrinsics.


kernels/sifive_x280/riscv_overloaded_intrinsics.h
Aaron-Hutchinson, line 1: Nitpick: Most of this file is organized into sections following the RVV spec. Some of the new definitions introduced here (`Integer operations`, `Mask operations`) don't appear to follow that organization.

Aaron-Hutchinson (commented):
Thanks for doing all of these conversions, @myeh01!

kernels/sifive_x280/riscv_overloaded_intrinsics.h
myeh01: Noted! I will fix this in a subsequent commit.

kernels/sifive_x280/1/bli_invscalv_sifive_x280_intr/bli_invscalv_sifive_x280_intr_complex.c
```diff
@@ -0,0 +1,83 @@
...
+        else
+            xvec = VLSSEG2_V_F(PREC, LMUL, 2)((BASE_DT*) x, 2 * FLT_SIZE * incx, vl);
+
+        RVV_TYPE_F(PREC, LMUL) xvec_real = VGET_V_F(PREC, LMUL, 2)(xvec, 0);
+        RVV_TYPE_F(PREC, LMUL) xvec_imag = VGET_V_F(PREC, LMUL, 2)(xvec, 1);
+        RVV_TYPE_F(PREC, LMUL) yvec_real, yvec_imag;
+
+        VCMUL_VF(PREC, LMUL, yvec_real, yvec_imag, xvec_real, xvec_imag, alpha_conj_inv.real, alpha_conj_inv.imag, vl);
+
+        RVV_TYPE_FX(PREC, LMUL, 2) yvec = VUNDEFINED_FX(PREC, LMUL, 2)();
```
leekillough, line 68: You are trying to assemble a complex vector in interleaved format. Based on my understanding, the counterpart to the desired `vssseg2e32.v` instruction are two `VSET_V_F()` intrinsics.

Shouldn't it be enough to just declare the variable?
```c++
RVV_TYPE_FX(PREC, LMUL, 2) yvec;
```

leekillough:
If you check the CI/CD build, you see one compiler error.

```
kernels/sifive_x280/1/bli_invscalv_sifive_x280_intr/./bli_invscalv_sifive_x280_intr_complex.c:68:43: fatal error: call to undeclared function '__riscv_vundefined_f32m4x2'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```

leekillough:
Integer types (`size_t`, `guint_t`, `int`) are mixed; some occurrences do not have a type conversion. This would be problem if potentially unsafe integer conversions were disallowed. However, I think that this is an issue throughout BLIS.

kernels/sifive_x280/1/bli_invscalv_sifive_x280_intr/bli_invscalv_sifive_x280_intr_complex.c
myeh01: If we only declare `yvec`, the compiler emits an uninitialized variable warning because `yvec` is called as an argument of `VSET_V_F` on L. 69. I bumped the toolchain again; the latest version should have `vundefined` intrinsics.

Aaron-Hutchinson: > the latest version should have `vundefined` intrinsics

In that case we should also fix the `#pragma` hack [that was used](https://github.com/flame/blis/blob/8820f8f91efd32e38e2995e73323656ef767bbd8/kernels/sifive_x280/1/bli_scal2v_sifive_x280_intr/bli_scal2v_sifive_x280_intr_complex.c#L82) in `scal2v` to ignore the warning. This is a different topic than this PR, so feel free to open a new PR afterwards if you prefer.

myeh01: I'll go ahead and make the change in this PR since it's a minor change if no one objects.

leekillough: No objection here

devinamatthews:
Wow, quite a lot of work. Thanks @myeh01. @fgvanzee unless you have any comment I'll merge this since everything is passing.

Issue #816, opened 2024-06-25
Not possible to link Blis and Lapack statically into the same executable


tbetcke:
I would like to link Blis and Lapack statically into the same executable. However, this is not possible since several symbols are multiply defined between Blis and Lapack. These are `lsame`, `xerbla`, `xerbla_array`, `crot`, `czymv`, `csyr`, `zrot`, ``zsymv`, `zsyr`. 

I noticed an earlier bug report #776 in which some of these were discussed and then included into Blis. Would it be possible to have an option to disable this in the future?  

The background is that I am integrating them into a Rust project. Rust prefers static linkage and does not handle rpath well for shared libraries.

fgvanzee:
We also just handled a similar request via #812. Your request, @tbetcke, has me wondering about a general-purpose solution that combines the compatibility accommodations for both 812 and your request into a single option, say, `--disable-symbols=<comma-separated-list-of-symbols>`. Then the user could omit whichever symbols they wish. (Of course, we'd have to guard the eligible symbols with cpp macros, but that's not difficult and it's something we can do incrementally as these offending symbols are brought to our attention.)

What do you think, @devinamatthews?

devinamatthews:
Yes, that's an interesting idea but it would be a lot of work to make it work "seamlessly" for end-users (e.g. they can disable any symbols within certain classes), and would be surprising to users if the list of symbols which can be disabled is small, irregularly defined, and/or constantly changing.

nick-knight:
As an author of applications that use BLIS, LAPACK, and ScaLAPACK together, I'm generally in favor of improving interoperability. Beyond resolving multiply defined symbols, what I'd really like is for BLIS implementations to override those in Sca/LAPACK. A partial solution would be for BLIS to name-mangle select routines with `__wrap_`, so I can use `--wrap` to prefer BLIS implementations. I believe that this solution is only partial because I don't see how it could hijack the libcalls inside the Sca/LAPACK binaries; achieving that I think would require build-system modifications in both BLIS and Sca/LAPACK. I don't know the best way forward to request such cross-project functionality.

rvdg:
I think the point Nick is making: the implementation in a BLAS library is likely to be higher performing than the reference implementation in LAPACK or ScaLAPACK (or some other library).  Hence, it is such a library that should selectively remove rather than, say, BLIS.

[...]

PR #823, opened 2024-08-01
Implemented `--omit-symbols=LIST` `configure` option.

fgvanzee:
Details:
- Added a new option to `configure` that allows the user to specify a list of symbols to omit from the library. The format of the option is `--omit-symbols=LIST` where `LIST` is a comma-separated list of symbol names (excluding any trailing underscore). This list is parsed into a list of `#define` directives that causes the relevant parts of BLIS to be ignored (or not). As such, the nature of this option is to only support omitting symbols which have been pre-identified as potential troublemakers when linking BLIS with other libraries such as LAPACK or ScaLAPACK. (This list may grow in the future as additional symbols are identified.)
- Re-implemented the `--enable-scalapack-compat` `configure` option to utilize the underlying `--omit-symbols=LIST` infrastructure.
- Implemented an `--enable-lapack-compat` option, which omits all of the known problematic symbols currently supported for omission.
- This commit addresses Issue #816. Thanks to Timo Betcke for bringing it to our attention and to Devin Matthews for his advice and for his initial implementation of `--enable-scalapack-compat` (PR #813).
- CREDITS file update.

cc @tbetcke

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -19,6 +19,7 @@ but many others have contributed code, ideas, and feedback, including
   Abhishek Bagusetty       @abagusetty                (Argonne National Laboratory)
   Satish Balay             @balay                     (Argonne National Laboratory)
   Kihiro Bando             @bandokihiro
+  Timo Betcke              @tbetcke                   (University College London)
   Matthew Brett            @matthew-brett             (University of Birmingham)
   Jérémie du Boisberranger @jeremiedbb
   Jed Brown                @jedbrown                  (Argonne National Laboratory)
--- a/build/bli_config.h.in
+++ b/build/bli_config.h.in
@@ -45,6 +45,9 @@
 // Enabled kernel sets (kernel_list)
 @kernel_list_defines@
 
+// Disabled symbols (symbol_omit_list)
+@omit_symbol_list_defines@
+
 #define BLIS_VERSION_STRING "@version@"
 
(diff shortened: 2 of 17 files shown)


fgvanzee:
I'm working on fixing some oversights in the original commit.

Separately, we'll need to decide whether omitting a symbol should also cause `blis.h` to omit its function prototype. Normally I'd say yes, but in the cases of `lsame_()` and `xerbla_()`, omitting prototypes causes a lot of warnings due to invocations of those functions the BLAS compatibility `_check()` functions. 🤔 (Naturally, if those symbols are omitted from the library, we won't be able to link against the library as-is, but it'd be nice if it at least *compiled* cleanly.)

devinamatthews:
> Separately, we'll need to decide whether omitting a symbol should also cause blis.h to omit its function prototype. Normally I'd say yes, but in the cases of lsame_() and xerbla_(), omitting prototypes causes a lot of warnings due to invocations of those functions the BLAS compatibility _check() functions.

I'd be OK making an exception to the "no symbol, no prototype" rule just for those two functions.

fgvanzee:
@tbetcke Since you were involved in #816, can you give this PR a try?

Take a look at `configure --help` to see the new symbol omission / compat options. Thanks for your help.

devinamatthews:
Good with me.

PR #827, opened 2024-09-20
Fix some aspects of the control tre/plugin infrastructure

devinamatthews:
1. Use configure-time variable substitution rather than the `PNAME` macro to generate symbol names in plugins. This makes it much easier for uses to see what names their symbols will have (and to change them if desired).
2. Use `siz_t` rather than `ukr_t` for anything dealing with kernel IDs (and similar for blocksizes and kernel preferences). Because users can now register new kernels, the values of the IDs for their custom kernels are no longer enumerated in `ukr_t`, which causes type conversion problems. This requires also being careful about the type of `BLIS_VA_END` and forcing existing enumerations like `ukr_t` to be represented using integers of the same width as `siz_t`.
3. Modify the `gemm` control tree initialization function to indicate whether or not the operation as a whole was transposed. This is needed if users have to treat the initial `A` and `B` differently in the control tree, for example in a tensor times matrix operation (if transposed to matrix times tensor, we need to know which "matrix" object is now the tensor).

### DIFF
--- a/build/plugin/bli_plugin.h.in
+++ b/build/plugin/bli_plugin.h.in
@@ -137,10 +137,10 @@ INSERT_GENTCONF
 #undef GENTCONF
 #define GENTCONF( CONFIG, config ) \
 \
-void PASTEMAC(plugin_init,BLIS_PNAME_INFIX,_,config)( PASTECH(plugin,BLIS_PNAME_INFIX,_params) ); \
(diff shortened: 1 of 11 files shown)


fgvanzee (approved):
Looks okay to me.

fgvanzee:
If you're ready, I'll format the commit log entry and commit.

PR #833, opened 2024-11-23
Optionally ignore extra dirs in `gen-make-frag.sh`.

fgvanzee:
Details:
- Implemented an option (`-i LIST`) to `gen-make-frag.sh` that allows the caller to optionally ignore additional directories when walking the source directory. (Note that previously the standard -- and only -- way to ignore directories was to add them to the `ignore_list` file, which is a required argument to the script.)
- I implemented this feature for something but then ended up not needing it, but figured it might be helpful in the future.

### DIFF
--- a/build/gen-make-frags/gen-make-frag.sh
+++ b/build/gen-make-frags/gen-make-frag.sh
@@ -86,6 +86,9 @@ print_usage()
 	echo "                 root_dir."
 	echo "   -h          hide"
 	echo "                 Hide the makefile fragments by prepending filenames with '.'."
+	echo "   -i LIST     ignore"
+	echo "                 Augment the list of directory names contained in ign_list"
+	echo "                 with the directory names in LIST."
 	echo "   -p PREFIX   prefix name"
 	echo "                 Use PREFIX instead of uppercased root_dir in the makefile"
 	echo "                 variable name. If the root_dir were 'stuff' and -p was not"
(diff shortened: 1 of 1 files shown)


build/gen-make-frags/gen-make-frag.sh
```diff
@@ -86,6 +86,9 @@ print_usage()
 	echo "                 root_dir."
 	echo "   -h          hide"
 	echo "                 Hide the makefile fragments by prepending filenames with '.'."
+	echo "   -i LIST     ignore"
+	echo "                 Augment the list of directory names contained in ign_list"
+	echo "                 with the directory names in LIST."
```
leekillough, line 91: Like `configure.sh`, here-docs can improve readability and maintainability.
```
cat <<EOF
Enter
Text
Here
EOF
```

fgvanzee: I agree. :)

build/gen-make-frags/gen-make-frag.sh
```diff
@@ -443,11 +447,12 @@ main()
 
 
 	# Process our command line options.
-	while getopts ":dho:p:rv:" opt; do
+	while getopts ":dhi:o:p:rv:" opt; do
 		case $opt in
 			d  ) dry_run_flag="1" ;;
 			h  ) hide_flag="1" ;;
 			r  ) recursive_flag="1" ;;
+			i  ) ignore_list="$OPTARG" ;;
```
devinamatthews, line 455: This should probably be `ignore_list="$ignore_list $OPTARG"` to allow for multiple -i options.

fgvanzee: Good idea. I've made the change.

Issue #825, opened 2024-08-28
OMP_NUM_THREADS discouraged? Why?


zerothi:
I am reading your wiki pages with comments on environment variables. Then I stumble upon:

https://github.com/flame/blis/blob/master/docs/Multithreading.md#environment-variables-the-automatic-way

Quote:

> Note: We highly discourage use of the OMP_NUM_THREADS environment variable to specify multithreading within BLIS and may remove support for it in the future. If you wish to set parallelism globally via environment variables, please use BLIS_NUM_THREADS.

Now, while I can see that having dedicated environment variables for the project, I think this goes largely against the end-user base.

At least from my perspective I do this:

1. The HPC admins compile and test the software and selects the appropriate libraries for performance.
2. The software compiled is made available for end-users, and hopefully one can pass generic information that lets them test and benchmark their code/use for their needs.

However, here comes the problem. When BLIS is compiled with OpenMP support, I would very much like to be able to tell our users about `OMP_NUM_THREADS`, as a first entry point to play with threading/parallelism.
This is the *de-facto* standard for end-users, and also the other related OMP flags, `OMP_PLACES`, `OMP_PROC_BIND`.

I would like the hierarchy to be simple:
```
BLIS_NUM_THREADS=X OMP_NUM_THREADS=Y  # use X
OMP_NUM_THREADS=Y  # use Y
BLIS_NUM_THREADS=X  # use X
```
In this way, users won't be troubled by *odd* naming conventions that means nothing to them.

PS. I know that fine-tuning the other BLIS specific env's is necessary, but most users won't bother with these details.

fgvanzee:
@zerothi Thanks for your feedback.

The following scenarios you provided
```
BLIS_NUM_THREADS=X OMP_NUM_THREADS=Y  # use X
OMP_NUM_THREADS=Y  # use Y
BLIS_NUM_THREADS=X  # use X
```
do indeed reflect how BLIS response to those environment variables. First, the library checks `BLIS_NUM_THREADS`; if it is set, that value is used. If it is not set, the library checks for `OMP_NUM_THREADS`; if it set, that value is used. If it is not set, then the number of total threads (not the concept, but the single user-facing number) is left unspecified. 

> Now, while I can see that having dedicated environment variables for the project, I think this goes largely against the end-user base.

[...]

PR #835, opened 2024-12-04
clarified OMP_NUM_THREADS

zerothi:
Removed description of possible deprecation of OMP_NUM_THREADS.

After discussion in #825 it was agreed to:

- lax the deprecation warning, OMP_NUM_THREADS will always be a fall-bock.
- Clarified with a simple example on how to do different regions of thread-counts.

### DIFF
--- a/docs/Multithreading.md
+++ b/docs/Multithreading.md
@@ -147,7 +147,14 @@ $ GOMP_CPU_AFFINITY="0-15" BLIS_NUM_THREADS=16 ./my_blis_program
 ```
 Either of these approaches causes BLIS to automatically determine a reasonable threading strategy based on what is known about the operation and problem size. If `BLIS_NUM_THREADS` is not set, BLIS will attempt to query the value of `BLIS_NT` (a shorthand alternative to `BLIS_NUM_THREADS`). If neither variable is defined, then BLIS will attempt to read `OMP_NUM_THREADS`. If none of these variables is set, the default number of threads is 1.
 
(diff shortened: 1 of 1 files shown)


zerothi:
What is not clear from the current document is how the `BLIS_IC_NT` and friends interact with the default number of threads. I didn't mention it currently.

I think this solves #825

devinamatthews:
Setting any of the `BLIS_XX_NT` values will override `BLIS_NUM_THREADS` and `OMP_NUM_THREADS`. Unset `BLIS_XX_NT` values are taken as 1. (@fgvanzee check me on this)

fgvanzee:
> What is not clear from the current document is how the `BLIS_IC_NT` and friends interact with the default number of threads. I didn't mention it currently.
> 
> I think this solves #825

This is actually already mentioned in the document under "Specifying multithreading." It's even in bold.

> Note: Please be aware of what happens if you try to specify both the automatic and manual ways, as it could otherwise confuse new users. Here are the important points:
>
> - Regardless of which of the three methods is used, **if multithreading is specified via both the automatic and manual ways, the values set via the manual way will always take precedence.**

zerothi:
Sorry for the mislead, I just didn't see it in the mentioned section. But I suppose that is not necessary, so please forget my comment.

I think the PR still remains final, let me know if you need any changes, I'll amend!

fgvanzee:
> Sorry for the mislead, I just didn't see it in the mentioned section. But I suppose that is not necessary, so please forget my comment.
> 
> I think the PR still remains final, let me know if you need any changes, I'll amend!

No problem. Thanks for your feedback!

PR #838, opened 2025-01-08
Fix errors and typos in some docs

jip:
Details:
- fix errors and typos,
- format according to coding style guide,
- unify a bit the formatting and content in `docs/BLIS*API.md`,
- sync the content in `docs/BLIS*API.md` with the current code base.

BTW, questions:
- There is one discrepancy in function signatures: `bli_?print[vm]` use the type `const void*` for `x` parameter, but `bli_?fprint[vm]` use the type `const ctype*`. Is this made intentionally?
- Why `bli_rntm_print` function is not exported?
- A `bli_info_get_enable_stay_auto_init` function is declared, but not defined.

PR #843, opened 2025-01-20
Fix errors and typos in some docs

jip:
Details:
- fix errors and typos,
- format according to coding style guide,
- unify a bit the formatting and content in `docs/BLIS*API.md`,
- sync the content in `docs/BLIS*API.md` with the current code base.

BTW, questions:
- There is one discrepancy in function signatures: `bli_?print[vm]` use the type `const void*` for `x` parameter, but `bli_?fprint[vm]` use the type `const ctype*`. Is this made intentionally?
- Why `bli_rntm_print` function is not exported?
- A `bli_info_get_enable_stay_auto_init` function is declared, but not defined.

Note: this is a refreshed version of my PR #838 closed that differs as follows:
- is rebased to the current master,
- commits are reordered a bit.

### DIFF
--- a/docs/BLISObjectAPI.md
+++ b/docs/BLISObjectAPI.md
@@ -74,15 +74,15 @@ The following tables list various types used throughout the BLIS object API.
 
 ### Integer-based types
 
-| BLIS integer type | Type definition          | Used to represent...                                                 |
-|:------------------|:-------------------------|:---------------------------------------------------------------------|
-| `gint_t`          | `int32_t` or `int64_t`   | general-purpose signed integer; used to define signed integer types. |
-| `guint_t`         | `uint32_t` or `uint64_t` | general-purpose signed integer; used to define signed integer types. |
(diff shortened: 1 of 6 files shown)


devinamatthews:
@jip thanks for rebasing. I'll check over the PR shortly. As for the questions:

- `bli_rntm_print` is an internal debugging function. You'll note it's not actually used in the framework either (calls are commented out).
- `void*` in `?print[vm]` is an oversight, and should really be `ctype*`. If you use `ctype` in the docs I can fix the code to match.
- AFAICT `bli_info_get_enable_stay_auto_init` is vestigial and the export can be removed. @fgvanzee?

jip:
>     * `void*` in `?print[vm]` is an oversight, and should really be `ctype*`. If you use `ctype` in the docs I can fix the code to match.

I've added a correcting commit for docs/BLISTypedAPI.md .

devinamatthews:
Thanks!

PR #844, opened 2025-01-23
Blacklist KNL with GCC 15+

loveshack:
This is necessary, but insufficient to build with (pre-release) GCC 15, which dropped the knl target.


devinamatthews:
The necessary but insufficient nature of this PR is noted, but by itself it looks great.

devinamatthews:
BTW when we come to a solution for #845 I will back-port this to 1.1 and 2.0.

Issue #848, opened 2025-01-24
lapack 3.12.1 compatibility with blis 1.1


cdluminate:
lapack 3.12.1 has introduced a couple of new symbols such as [`sgemmtr_`](https://github.com/Reference-LAPACK/lapack/pull/887). But it seems missing from BLIS 1.1. As described in the pull request, when we use `update-alternatives` to switch to BLIS, the program will fail to find that symbol, see example [here](https://ci.debian.net/packages/p/pytorch/unstable/amd64/56757882/). Some relevant test failures from the Debian side also points to the same issue: https://qa.debian.org/excuses.php?experimental=1&package=lapack .

Any suggestion on this? Is copying some code from lapack and glue them into BLIS a reasonable temporary workaround?

devinamatthews:
`?gemmtr` is just `bli_gemmt` so it should be easy to add an interface adapter. If you are willing to work on a PR then I suggest taking an existing level-3 BLAS interface (`gemm`, `syrk`, etc.) and adapting it.

cdluminate:
Thanks for the pointer. I'll take a look

cdluminate:
I noted that there is already `frame/compat/extra/bla_gemmt.h`, and a symbol `sgemmt_` in the compat shared object. However, the upstream does not provide that `sgemmt_`. Should we instead rename the `gemmt` into `gemmtr` in the compat code?

```
$ readelf -sW /usr/lib/x86_64-linux-gnu/blis-openmp/libblas.so.3 | grep sgemm
    75: 0000000000805d50   831 FUNC    GLOBAL DEFAULT   12 cblas_sgemmt
   132: 0000000000807b40  2485 FUNC    GLOBAL DEFAULT   12 sgemm_batch_
   136: 00000000007fc470   742 FUNC    GLOBAL DEFAULT   12 cblas_sgemm
   234: 0000000000805950  1013 FUNC    GLOBAL DEFAULT   12 cblas_sgemm_batch
   249: 00000000007e2a00  2367 FUNC    GLOBAL DEFAULT   12 sgemm_
   380: 000000000080a290  1669 FUNC    GLOBAL DEFAULT   12 sgemmt_
```


```
$ readelf -sW tmp/usr/lib/x86_64-linux-gnu/blas/libblas.so.3.12.1 | grep sgemm
    24: 0000000000047d80  2941 FUNC    GLOBAL DEFAULT   12 sgemmtr_
[...]

PR #849, opened 2025-01-28
Alias *gemmt_ as *gemmtr_ to fix lapack 3.12.1 compatibility. (Closes: #848)

cdluminate:
The patch is sufficient to address the issue that `liblapack.so` says it cannot find symbol `sgemmtr_`. The `cblas` part is not yet done. I'll look into that and update this PR. Or if it works for you, I can do a separate PR for the `cblas` alias.


cdluminate:
@devinamatthews I pushed the changes for `cblas_?gemmtr` as well. The PR should be able to fully address #848.

PR #859, opened 2025-02-27
Fix check for SVE instructions which caused problems on Windows.

### LINKED ISSUE #858: 1.1: ARMSVE fails to build on Windows
The runtime detection of the presence of the CPU feature uses a method not available in Windows.
```
Compiling obj/arm64/kernels/armsve/1m/bli_dpackm_armsve256_int_8xk.o ('armsve' CFLAGS for kernels)
../blis-1.1/config/armsve/bli_cntx_init_armsve.c:36:10: fatal error: 'sys/auxv.h' file not found
   36 | #include <sys/auxv.h>
      |          ^~~~~~~~~~~~
1 error generated.
make: *** [Makefile:707: obj/arm64/config/armsve/bli_cntx_init_armsve.o] Error 1
make: *** Waiting for unfinished jobs....
==> ERROR: A failure occurred in build().
```
A Windows compatible way of finding the same needs to be added.

BLIS 1.1

devinamatthews:
Details:
- The context intialization for `armsve` was using the HWCAP functionality of Linux to check if SVE instructions are actually available, since these are used to determine the register blocksizes. Naturally, this causes problems on Windows.
- Instead, use functions from `bli_cpuid.c` to check for SVE. On Windows, no check is actually done and SVE is never detected.
- In the case that (sometime in the future) the user specifically requests the `armsve` config on Windows, only enable this check for the whole `arm64` family and just assume SVE is available otherwise.
- Fixes #858.
- `armsve` has been blacklisted for Windows due to other compilation issues.

### DIFF
--- a/config/armsve/bli_cntx_init_armsve.c
+++ b/config/armsve/bli_cntx_init_armsve.c
@@ -33,21 +33,24 @@
 */
 
 #include "blis.h"
-#include <sys/auxv.h>
-
-#ifndef HWCAP_SVE
-#define HWCAP_SVE (1 << 22)
-#endif
 
 void bli_cntx_init_armsve( cntx_t* cntx )
 {
-	if (!(getauxval( AT_HWCAP ) & HWCAP_SVE))
-		return;
-
-	blksz_t blkszs[ BLIS_NUM_BLKSZS ];
-
 	// Set default kernel blocksizes and functions.
 	bli_cntx_init_armsve_ref( cntx );
+
+	// If we are autodetecting the correct aarch64 config, then we have to make sure
+	// that SVE instructions are actually available since these are used in determining
+	// the register blocksizes.
+	#ifdef BLIS_FAMILY_ARM64
+	uint32_t family, model, features = 0;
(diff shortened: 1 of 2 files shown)


config/armsve/bli_cntx_init_armsve.c
```diff
@@ -33,22 +33,25 @@
...
-	if (!(getauxval( AT_HWCAP ) & HWCAP_SVE))
+	// Set default kernel blocksizes and functions.
+	bli_cntx_init_armsve_ref( cntx );
+
+	// If we are autodetecting the correct aarch64 config, then we have to make sure
+	// that SVE instructions are actually available since these are used in determining
+	// the register blocksizes.
+	#ifdef BLIS_FAMILY_ARM64
+	uint32_t family, model, features = 0;
+	bli_cpuid_query( &family, &model, &features );
```
hmartinez82, line 47: I'm going to try to apply this patch, but I'm almost sure it won't work. ARM64 doesn't have cpuid, and the source code for `bli_cpuid_query()` seems to be totally made for x86 only.

Issue #864, opened 2025-04-06
crazy bug: oapi/09level3.x and tapi/04level3.x


masonmcbride:
### Example 1: blis/examples/tapi/04level3.x
https://cdn.discordapp.com/attachments/879093420801527861/1358475995321077780/image.png?ex=67f3fadc&is=67f2a95c&hm=88235bc217a43c647ddba94e64ab9ffc3c6754c618599f5470735ee08ca9ea54&
https://cdn.discordapp.com/attachments/879093420801527861/1358476722735157540/image.png?ex=67f3fb89&is=67f2aa09&hm=3be08158dd8ebd4b326ed0dc155c6fd2fc0afefc10b5a4aa223e8504d975ea7b&
Both of these examples have incorrect computations of the dot product. If a value is multiplied by zero it becomes one (instead of zero) and makes the computation incorrect. 

### Example 2: blis/examples/oapi/09level3.x 
https://cdn.discordapp.com/attachments/879093420801527861/1358481591713923364/image.png?ex=67f40012&is=67f2ae92&hm=98b076f1d9796c714b3f034023039ce04264279b3187faaa76bf54982141dc6b&
Oapi still has the same issue. 

`BLIS_ARCH_TYPE=3 ./09level3.x` 
### Info
version: 53d21cb4
architecture: Haswell
os: windows 11 using WSL 2 , with an intel i12700k. 
=== Versions ===
gcc (Ubuntu 13.1.0-8ubuntu1~22.04) 13.1.0
GNU ld (GNU Binutils for Ubuntu) 2.38
GNU assembler (GNU Binutils for Ubuntu) 2.38
GNU ar (GNU Binutils for Ubuntu) 2.38
GNU Make 4.3

hominhquan:
Thanks @masonmcbride for noticing this. It turns out that the floating-point format string used in `bli_printm` might not be precise enough (`4.1f`). If it is set to something like `4.3f`, the output is much more mathematically correct.

I'll push a PR for this.

devinamatthews:
Thank you @hominhquan of course it is the simple and obvious thing that I didn't think of :).

PR #865, opened 2025-04-06
examples: replace all 4.1f printm format by 4.3f

hominhquan:
- This avoids possible mis-interpretation of computation results printed on stdout (thanks Mason McBride for reporting it in #864)


devinamatthews:
Would not compile in Fortran, but luckily we don't have to code in Fortran. Thanks @hominhquan.

PR #895, opened 2025-09-18
Optimize bli_gemm_haswell_asm_d6x8: reduce latency & improve throughput

harsdave:
- Reorder instructions to reduce pipeline stalls and dependencies
- Adjust register usage to improve data reuse
- tweak prefetcch so cache warming is more effective


harsdave:
> Because you've added general stride support to the assembly region, you should change `GEMM_UKR_SETUP_CT_AMBI` to `GEMM_UKR_SETUP_CT_ANY`. Otherwise, the C version will still be used for general stride.

Corrected, thank you!

PR #918, opened 2026-02-24
add cpu detection for intel scalable thru emerald rapids and xeon 6

### LINKED ISSUE #917: Emerald rapids detected as haswell
Problem: the emeraId rapids 8562Y is detected as 'haswell' (the output of auto-detect.x).
This CPU has AVX-512 and I'm not sure why this is happening.
Compiler: GCC 13.3.0
commit: b5d57831a9cecd6e7e83b039be66fbf31a005062

EDIT:
With logging enabled, the reason is
`libblis: Number of FMA units unknown; using 'haswell' (not 'skx') config`
because there is no data for this processor in vpu_count().

chillenb:
This is an attempt to add the scalable Xeon chips through Emerald Rapids to bli_cpuid.h in order to fix #917.
Some of the silver models in newer generations have 2 vpus, so I had to adjust the logic a little.

I also tried to add the Xeon 6 series, which all have 2 vpus except for 6300--6399. I'm assuming that the model strings don't have a separate "6" in them, i.e. "Intel(R) Xeon(R) 6740E Processor" and not "Intel(R) Xeon(R) 6 6740E Processor". These are distinguished from gold CPUs with overlapping sku ranges by the absence of precious metals.

### DIFF
--- a/frame/base/bli_cpuid.c
+++ b/frame/base/bli_cpuid.c
@@ -987,6 +987,34 @@ int vpu_count( void )
 		else if ( 2199 >= sku && sku >= 2120 ) return 2;
 		else if ( 2102 == sku || sku == 2104 ) return 2; // Gold exceptions
 		else if ( 2119 >= sku && sku >= 2100 ) return 1;
+
+		// 3rd gen Intel Scalable and W all have 2 vpus.
+		else if ( 8399 >= sku && sku >= 8300 ) return 2; // Ice Lake Platinum
+		else if ( 6399 >= sku && sku >= 6300 ) return 2; // Ice Lake Gold
+		else if ( 5399 >= sku && sku >= 5300 ) return 2; // Ice Lake Gold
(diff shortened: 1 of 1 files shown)


chillenb:
It might be good to return 2 for unknown Xeon cpus with avx512, and add exceptions over time, rather than return -1 by default.

jeffhammond:
Agree on defaulting to 2. I drove the decision to have 2 in all ICX SKUs and it seemed like this would hold in future products, although I left the company in 2021 and ceased to have any involvement in this topic.

jeffhammond:
Cross reference with https://github.com/jeffhammond/vpu-count. 5122 has 2 FMAs in it, not that it really matters.

chillenb:
Thanks for your help!

(fortunately the 5122 was already handled on line 988)
If you still worked for Intel, I'd beg you for a fully 8-wide divider/sqrt. Sigh

Issue #780, opened 2023-10-10
triggers segfault in "make check" process with configuration --disable-sba-pools on CentOS (SkylakeX)


realab-ai:
While config the master branch (commit 1e264a4) with "./configure --enable-threading=openmp --disable-sba-pools auto" and compile with gcc/11.2.0, there is a segment error in "make check" process:
```bash
Running test_libblis.x (fast) with output redirected to 'output.testsuite'
Running cblat1.x > 'out.cblat1'
Compiling obj/skx/blastest/cblat2.o
Compiling obj/skx/blastest/cblat3.o
Running dblat1.x > 'out.dblat1'
Compiling obj/skx/blastest/dblat2.o
Compiling obj/skx/blastest/dblat3.o
Running sblat1.x > 'out.sblat1'
Compiling obj/skx/blastest/sblat2.o
Compiling obj/skx/blastest/sblat3.o
Running zblat1.x > 'out.zblat1'
Compiling obj/skx/blastest/zblat2.o
Compiling obj/skx/blastest/zblat3.o
bash: line 2:   693 Segmentation fault      ./test_libblis.x -g ./testsuite/input.general.fast -o ./testsuite/input.operations.fast > output.testsuite
make: *** [testsuite-run-fast] Error 139
make: *** Waiting for unfinished jobs....
rm obj/skx/blastest/dblat3.o obj/skx/blastest/zblat2.o obj/skx/blastest/cblat2.o obj/skx/blastest/sblat2.o obj/skx/blastest/cblat3.o obj/skx/blastest/dblat2.o obj/skx/blastest/sblat3.o obj/skx/blastest/zblat3.o
```

chillenb:
Could be caused by #919. Try checking with --enable-asan.

PR #920, opened 2026-02-24
ensure thrinfo is valid at the end of bli_l3_sup_thread_decorator_entry

### LINKED ISSUE #919: Use after free in bli_gemmsup_int
In bli_l3_sup_thread_decorator_entry,
https://github.com/flame/blis/blob/b5d57831a9cecd6e7e83b039be66fbf31a005062/frame/3/bli_l3_decor.c#L70-L90
thrinfo_t thread gets passed into `bli_gemmsup_int`, which frees it via `bli_l3_sup_thrinfo_update`.

Line 88 bli_l3_sup_thread_decorator_entry then tries to call `bli_thrinfo_barrier( thread ); ` using the previously freed `thrinfo_t thread`.
This causes tests to fail when mimalloc is preloaded and pba/sba are disabled, since `thread` no longer contains valid data.

One solution is to change the signature of `bli_gemmsup_int` like so
```c
 err_t bli_gemmsup_int
     (
       const obj_t*  alpha,
       const obj_t*  a,
       const obj_t*  b,
       const obj_t*  beta,
       const obj_t*  c,
       const cntx_t* cntx,
       const rntm_t* rntm,
             thrinfo_t** thread
     )
```
This function should then return a valid thrinfo_t.

chillenb:
This is an attempt to fix #919. I'm not too familiar with the codebase, so I'm not sure if the barrier at the end of `bli_l3_sup_thread_decorator_entry` needs to use the _original_ thrinfo or after modification inside bli_gemmsup_int / bli_gemmtsup_int.

### DIFF
--- a/frame/3/bli_l3_sup_int.c
+++ b/frame/3/bli_l3_sup_int.c
@@ -139,7 +139,7 @@ err_t bli_gemmsup_int
 			// new ways of parallelism value for the jc loop.
 			rntm_t rntm_l = *rntm;
 			bli_rntm_set_ways_only( jc_new, 1, ic_new, 1, 1, &rntm_l );
-			bli_l3_sup_thrinfo_update( &rntm_l, &thread );
+			bli_l3_sup_thrinfo_update( &rntm_l, thread );
 		}
 
 
@@ -205,7 +205,7 @@ err_t bli_gemmsup_int
 			// new ways of parallelism value for the jc loop.
 			rntm_t rntm_l = *rntm;
 			bli_rntm_set_ways_only( jc_new, 1, ic_new, 1, 1, &rntm_l );
-			bli_l3_sup_thrinfo_update( &rntm_l, &thread );
+			bli_l3_sup_thrinfo_update( &rntm_l, thread );
 		}
 
 
@@ -315,7 +315,7 @@ err_t bli_gemmtsup_int
 			// new ways of parallelism value for the jc loop.
 			rntm_t rntm_l = *rntm;
 			bli_rntm_set_ways_only( jc_new, 1, ic_new, 1, 1, &rntm_l );
-			bli_l3_sup_thrinfo_update( &rntm_l, &thread );
+			bli_l3_sup_thrinfo_update( &rntm_l, thread );
 		}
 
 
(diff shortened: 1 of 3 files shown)


chillenb:
I decided to do it a different way (less invasive, fewer changes)

1. Ensure bli_l3_sup_thrinfo_update(rntm, root) doesn't free root. Instead, it frees and rebuilds the subtree below root.
2. Change the signature
```c
void bli_l3_sup_thrinfo_update
     (
       const rntm_t*     rntm,
             thrinfo_t** root
     );
```
to
```c
void bli_l3_sup_thrinfo_update
     (
       const rntm_t*     rntm,
             thrinfo_t*  root
     );
```
to reflect the fact that `root` is not being freed.

EylonKrause:
I reproduced #919 on a Zen4 (Ryzen 9 8940HX, Ubuntu 24.04) and can confirm this PR fixes it. A few notes, since it's been waiting a while:

**The UAF (confirmed):** `bli_l3_sup_thrinfo_update()` frees the root `thrinfo_t` and returns a replacement via the `thrinfo_t**` out-param, updating only the local `thread` in `bli_gemmsup_int()`. The decorator `bli_l3_sup_thread_decorator_entry()` keeps the original pointer and uses it for the closing `bli_thrinfo_barrier()` / `bli_thrinfo_free()`. With `--enable-asan --disable-sba-pools`:

```
==ERROR: AddressSanitizer: heap-use-after-free   READ of size 8
  #0 bli_l3_sup_thread_decorator_entry
freed by: bli_l3_sup_thrinfo_update
```

**This PR fixes it, and also fixes #780.** With this PR applied, the fast testsuite under `--enable-asan --disable-sba-pools --disable-pba-pools` with `BLIS_NUM_THREADS=4` runs with **0 ASan errors, all PASS**. #780 (the `make check` segfault under `--disable-sba-pools`) is the same root cause and is resolved by this change. In the shipping config it's 2772/2772 testsuite PASS and all BLAS drivers pass, with no numerical regression.

One small suggestion: the rebuilt subtree logic in `bli_l3_sup_thrinfo_update()` duplicates the block in `bli_l3_sup_thrinfo_create()`. It could be factored into a shared helper, mirroring the existing `bli_l3_thrinfo_create()` / `bli_l3_thrinfo_grow()` split. I put up #931 as an alternative that does exactly that, but the approach here is sound either way — whichever you prefer.

devinamatthews:
@chillenb I'm merging #931 as a slight tweak of this. Thanks to both you and @EylonKrause.

PR #924, opened 2026-03-24
Implement optimized ccopyv and zcopyv kernels for Zen/2/3

harsdave:
Description:
This patch implements high-performance complex copy (ccopyv) and double-complex copy (zcopyv) kernels.

Key Changes:

Vectorization: Utilizes AVX/AVX2 intrinsics (__m256, __m256d) to process multiple complex elements per cycle for unit-strided (incx == 1, incy == 1) cases.

Conjugation Support: Implements efficient on-the-fly conjugation for bli_is_conj cases using sign-flip masks (_mm256_setr_ps(1, -1, ...)), avoiding separate passes.

Loop Unrolling: Employs an 8-register unrolling scheme (32 elements for ccopyv, 16 for zcopyv) to maximize instruction-level parallelism and hide memory latency.

### DIFF
--- a/kernels/zen/1/bli_copyv_zen_int.c
+++ b/kernels/zen/1/bli_copyv_zen_int.c
@@ -334,3 +334,533 @@ void bli_dcopyv_zen_int
 	}
 }
 
+void bli_ccopyv_zen_int
+(
+             conj_t  conjx,
+             dim_t   n,
+       const void*   x, inc_t incx,
+             void*   y, inc_t incy,
+       const cntx_t* cntx
+)
+{
+	const scomplex* x0 = x;
+	      scomplex* y0 = y;
+	// If the vector dimension is zero return early.
+	if ( bli_zero_dim1( n ) )
+	{
+		return;
+	}
+
+  // Setting the local pointers and iterator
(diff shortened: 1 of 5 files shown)


kernels/zen/1/bli_copyv_zen_int.c
```diff
@@ -334,3 +334,533 @@ void bli_dcopyv_zen_int
...
+				y0 += n_elem_per_reg;
+			}
+
+		}
+
+		// Handling fringe cases or non-unit strided inputs
+		for (; i < n; i += 1)
+		{
+			scomplex temp = *x0;
+      temp.imag = -temp.imag;
```
devinamatthews, line 460: Does this work if scomplex is defined as C99 _Complex float?

devinamatthews (changes requested):
Looks like I also missed the conjugation issue [here](https://github.com/flame/blis/commit/c30230e256dfb302225219a225eb443eaca024aa#diff-dbbe4000e9f30694e78a5920afc75d7bfc3237f943fa67f68c307a6346e21dcaR257). Please fix (in this PR is OK) if needed.

PR #925, opened 2026-03-26
Amd zen4 support

harsdave:
This patch introduces the 'zen4' configuration.

Key Changes:
- Added 'zen4' configuration directory and base make_defs.mk.
- Implemented an optimized daddv kernel (bli_daddv_zen4_int) using
  AVX-512 intrinsics.
- The daddv implementation utilizes:
    * 8x/4x/2x unrolling for unit-stride vectors to maximize FMA throughput.
    * AVX-512 masked loads/stores for tail (fringe) cases, eliminating
       the need for scalar fallback loops for non-unit multiples.
 - Initial configuration uses 'zen' fallbacks for remaining Level-1
    kernels, which are scheduled for AVX-512 optimization in future updates.


devinamatthews (changes requested):
Can you please add zen2/3/4 tests to `ci/do_sde.sh`? You will need to grab CPUID dumps from relevant machines (or fabricate them).

PR #927, opened 2026-03-30
Automatically detect hardware type for zen5 and later

edwsmyth:
Add fallback code based on detecting ISA support to allow AMD processors not specifically supported yet to identify as the latest supported Zen sub-config.

### DIFF
--- a/frame/base/bli_cpuid.c
+++ b/frame/base/bli_cpuid.c
@@ -5,7 +5,7 @@
    libraries.
 
    Copyright (C) 2014, The University of Texas at Austin
-   Copyright (C) 2018-2020, Advanced Micro Devices, Inc.
+   Copyright (C) 2018 - 2026, Advanced Micro Devices, Inc. All rights reserved.
    Copyright (C) 2019, Dave Love, University of Manchester
 
    Redistribution and use in source and binary forms, with or without
@@ -82,6 +82,18 @@
 
 #include "cpuid.h"
 
+// Variables to return if different AVX instructions are supported. These will
+// be updated in bli_cpuid_query_id(), which is called only once from
+// bli_arch_set_id(), using the pthread_once mechanism. This assume that the
+// cpuid information will not change once the library is loaded.
+// Cached values can be returned by the specific functions for each below.
+
+static bool is_avx2fma3_supported = FALSE;
+static bool is_avx512_supported = FALSE;
+static bool is_avx512vnni_supported = FALSE;
+static bool is_avx512bf16_supported = FALSE;
+static bool is_avx512fp16_supported = FALSE;
+
 arch_t bli_cpuid_query_id( void )
 {
 	uint32_t vendor, family, model, features;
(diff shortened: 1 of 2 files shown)


frame/base/bli_cpuid.c
```diff
@@ -485,6 +522,204 @@ bool bli_cpuid_is_bulldozer
...
+// Determine if the CPU has support for AVX2 and FMA3.
+void bli_cpuid_check_avx2fma3_support
+     (
+       uint32_t family,
+       uint32_t model,
+       uint32_t features
+     )
+{
+	// Check for expected CPU features.
+	const uint32_t expected = FEATURE_AVX     |
```
devinamatthews, line 534: It would be better not to have static global variables. Instead you could do:
```c
enum
{
    FEATURE_SET_AVX2FMA3 = this | that | theother,
    ...
};
```
and use `bli_cpuid_has_features` directly thereafter.

frame/base/bli_cpuid.c
```diff
@@ -485,6 +522,204 @@ bool bli_cpuid_is_bulldozer
...
+// support routines below.
+
+static bli_pthread_once_t once_check_cpuid_query_id = BLIS_PTHREAD_ONCE_INIT;
+
+void bli_cpuid_query_id_wrapper( void )
+{
+	arch_t __attribute__ ((unused)) id_w;
+        id_w = bli_cpuid_query_id();
+}
+void bli_cpuid_query_id_once( void )
```
devinamatthews, line 681: If you want to avoid repeated overhead, better to cache return value of `bli_cpuid_query_id` (perhaps rename to `bli_cpuid_query_id_impl`) in static global variable of `bli_cpuid_query_id_once` (rename without `_once`to avoid changing lots of dependents). Again not crazy about global vars but just one for caching is OK. This does of course break big.LITTLE or P/E systems but there are larger issues there anyways.

frame/base/bli_cpuid.c
```diff
@@ -144,6 +175,12 @@ arch_t bli_cpuid_query_id( void )
 		if ( bli_cpuid_is_zen( family, model, features ) )
 			return BLIS_ARCH_ZEN;
 #endif
+#ifdef BLIS_CONFIG_ZEN3
+		// Fallback test for future AMD processors
+		// Use zen3 if AVX512 support is not available but AVX2 is.
+		if ( is_avx2fma3_supported )
```
devinamatthews, line 181: No AVX512 test for Zen4/5 fallback?

frame/base/bli_cpuid.c
```diff
@@ -485,6 +522,204 @@ bool bli_cpuid_is_bulldozer
...
+}
+void bli_cpuid_query_id_once( void )
+{
+#ifndef BLIS_CONFIGURETIME_CPUID
+	bli_pthread_once( &once_check_cpuid_query_id, bli_cpuid_query_id_wrapper );
+#endif
+}
+
+// API to check if AVX2 and FMA3 are supported or not on the current platform.
+bool bli_cpuid_is_avx2fma3_supported( void )
```
devinamatthews, line 689: Are these actually needed? They're not used ATM.

devinamatthews (changes requested):

PR #931, opened 2026-06-24
Fix use-after-free of root thrinfo_t in sup thread decorator

### LINKED ISSUE #919: Use after free in bli_gemmsup_int

EylonKrause:
## Summary

`bli_l3_sup_thrinfo_update()` freed the **root** `thrinfo_t` node and returned a replacement through its `thrinfo_t**` out-parameter, updating only the local `thread` pointer inside `bli_gemmsup_int()` / `bli_gemmtsup_int()`. The caller `bli_l3_sup_thread_decorator_entry()` (`frame/3/bli_l3_sup_decor.c`) keeps its own copy of that pointer and uses it after the call for the closing `bli_thrinfo_barrier()` and `bli_thrinfo_free()`, so it reads and frees already-freed memory — a heap-use-after-free followed by a double-free of the root node.

Fixes #919 and resolves the long-standing testsuite crash in #780.

## Why it was hidden

BLIS's `sba`/`pba` pools recycle the freed block, so normal builds appear fine. It becomes observable when the pools are disabled (`./configure --disable-sba-pools`, exactly the configuration reported in #780) or under AddressSanitizer (`./configure --enable-asan`).

## Fix

Re-factorize the thread control tree **in place**: keep the root node (which the decorator still owns) and free + regrow only the subtree beneath it. The shared subtree construction is factored into a new `bli_l3_sup_thrinfo_grow()` helper, mirroring the existing `bli_l3_thrinfo_create()` / `bli_l3_thrinfo_grow()` split, so `create` and `update` no longer duplicate it. `bli_l3_sup_thrinfo_update()` now takes `thrinfo_t*` instead of `thrinfo_t**`.

## Credit / relation to #920

The diagnosis and the root-preserving strategy are due to @chillenb (#919, #920). This PR implements the same strategy but factors the duplicated subtree construction into a shared helper. It is offered as an alternative to #920 — happy to defer to whichever the maintainers prefer.

## Verification (Zen4, Ryzen 9 8940HX, Ubuntu 24.04)

**Before** — `./configure --enable-asan --disable-sba-pools --disable-pba-pools auto`, fast testsuite, `BLIS_NUM_THREADS=4`:

```
==ERROR: AddressSanitizer: heap-use-after-free   READ of size 8   thread T2
  #0 bli_l3_sup_thread_decorator_entry
freed by thread T2 here:
  #1 bli_l3_sup_thrinfo_update
SUMMARY: AddressSanitizer: heap-use-after-free in bli_l3_sup_thread_decorator_entry
```

**After** — same ASan + pools-disabled config: **0 ASan errors, all PASS, exits normally.**

**Shipping config** — `./configure auto` (pools enabled), `BLIS_NUM_THREADS=4`:
- `make checkblis-fast`: **2772 / 2772 PASS**
- `make checkblas`: **All BLAS tests passed!**

No numerical regression.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -71,6 +71,7 @@ but many others have contributed code, ideas, and feedback, including
   Mike Kistler             @mkistler                  (IBM, Austin Research Laboratory)
   Nick Knight              @nick-knight               (SiFive)
   Ivan Korostelev          @ivan23kor                 (University of Alberta)
+  Eylon Krause             @EylonKrause
   Kyungmin Lee             @kyungminlee               (Ohio State University)
   Michael Lehn             @michael-lehn
                            @leo4678
--- a/frame/3/bli_l3_sup_int.c
+++ b/frame/3/bli_l3_sup_int.c
@@ -139,7 +139,7 @@ err_t bli_gemmsup_int
 			// new ways of parallelism value for the jc loop.
 			rntm_t rntm_l = *rntm;
 			bli_rntm_set_ways_only( jc_new, 1, ic_new, 1, 1, &rntm_l );
-			bli_l3_sup_thrinfo_update( &rntm_l, &thread );
+			bli_l3_sup_thrinfo_update( &rntm_l, thread );
 		}
 
 
@@ -205,7 +205,7 @@ err_t bli_gemmsup_int
 			// new ways of parallelism value for the jc loop.
 			rntm_t rntm_l = *rntm;
 			bli_rntm_set_ways_only( jc_new, 1, ic_new, 1, 1, &rntm_l );
-			bli_l3_sup_thrinfo_update( &rntm_l, &thread );
+			bli_l3_sup_thrinfo_update( &rntm_l, thread );
 		}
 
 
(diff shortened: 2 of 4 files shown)


devinamatthews:
Excellent work! Thanks for the fix.

PR #932, opened 2026-06-24
Accumulate single-precision BLAS norms (snrm2, scnrm2) in double

### LINKED ISSUE #914: SNRM2 precision
There is a thread on OpenBlas
https://github.com/OpenMathLib/OpenBLAS/issues/5503

which may be of some interest for you guys as well. 

Currently, out of NETLIB, BLIS, OpenBlas, Intel MKL, Apple Veclib, 
NETLIB and BLIS have low (but consistent with `single` precision), 
and others have precision higher than eps(single) (assumingly by doing
summation in `double` precision). 
I think it would be great if BLIS would do the same, but as a minimum
am curios about your opinion on this matter.

EylonKrause:
## Summary

The BLAS single-precision norms `snrm2` and `scnrm2` route through `bli_?normfv`, which accumulates the sum of the squares in the working (single) precision. Their results are therefore noticeably less accurate than other BLAS implementations (Intel MKL, Apple vecLib, OpenBLAS), which accumulate in double. This addresses #914.

This PR accumulates the sum of the squares in **double** for the single-precision norms only, in the BLAS interface (`frame/compat/bla_nrm2.c`). The double-precision norms (`dnrm2`, `dznrm2`) are unchanged and keep using BLIS's scaled sum-of-squares kernel, which guards against unnecessary overflow. Overflow is not a concern for the single→double path: a single-precision magnitude squared is at most ~1.16e77, far below `DBL_MAX` (~1.8e308).

## Demonstration

Vector `x = [30000, then 4,000,000 ones]` (exact sum of squares 9.04e8, true norm ≈ 30066.5928):

| | `snrm2` | rel. error vs. long-double reference |
|---|---:|---:|
| master  | 30000.0000 | 2.2e-03  (the unit contributions vanish in float accumulation) |
| this PR | 30066.5918 | 3.2e-08  (≈ single-precision epsilon) |

## Validation (Zen4, Ryzen 9 8940HX, Ubuntu 24.04)

- `make checkblas` — BLAS reference drivers including `sblat1` (which tests `snrm2`): **All BLAS tests passed!**
- `make checkblis-fast` — BLIS testsuite norm operations: **2772 / 2772 PASS**

## Notes

This fixes the **BLAS-level** `snrm2`/`scnrm2` (and `cblas_snrm2`/`cblas_scnrm2`, which route through them). The native `bli_?normfv` kernel is left unchanged; I'm happy to move the higher-precision accumulation into the kernel instead if you'd prefer it to apply to the native API as well.

Closes #914.

### DIFF
--- a/configure
+++ b/configure
@@ -287,6 +287,14 @@ print_usage()
                  Enabling this option WILL NEGATIVELY IMPACT PERFORMANCE.
                  Please use only for informational/debugging purposes.
 
+   --disable-mixed-precision-norm, --enable-mixed-precision-norm
+
+                 Disable (enabled by default) accumulating the single-precision
+                 vector norms (snrm2, scnrm2) in double precision. When enabled
+                 (the default) the results are consistent with other BLAS
+                 implementations; disabling restores accumulation in single
+                 precision.
+
    --enable-asan, --disable-asan
 
                  Enable (disabled by default) compiling and linking BLIS
@@ -3056,6 +3064,7 @@ blis_main()
 	enable_pba_pools='yes'
 	enable_sba_pools='yes'
 	enable_mem_tracing='no'
+	enable_mixed_precision_norm='yes'
 	int_type_size=0
 	blas_int_type_size=32
 	enable_blas='yes'
@@ -3267,6 +3276,12 @@ blis_main()
 							;;
 						disable-mem-tracing)
 							enable_mem_tracing='no'
+							;;
+						enable-mixed-precision-norm)
+							enable_mixed_precision_norm='yes'
+							;;
+						disable-mixed-precision-norm)
+							enable_mixed_precision_norm='no'
 							;;
 
 						enable-addon=*)
@@ -4059,6 +4074,13 @@ blis_main()
 		echo "${script_name}: memory tracing output is disabled."
 		enable_mem_tracing_01=0
 	fi
+	if [[ ${enable_mixed_precision_norm} = yes ]]; then
+		echo "${script_name}: double-accumulated single-precision norms (snrm2, scnrm2) are enabled."
+		enable_mixed_precision_norm_01=1
+	else
+		echo "${script_name}: double-accumulated single-precision norms (snrm2, scnrm2) are disabled."
+		enable_mixed_precision_norm_01=0
+	fi
 	if [[ ${has_memkind} = yes ]]; then
 		if [[ -z ${enable_memkind} ]]; then
 			# If no explicit option was given for libmemkind one way or the other,
@@ -4488,6 +4510,7 @@ blis_main()
 	add_config_var enable_pba_pools          enable_pba_pools_01
 	add_config_var enable_sba_pools          enable_sba_pools_01
 	add_config_var enable_mem_tracing        enable_mem_tracing_01
+	add_config_var enable_mixed_precision_norm enable_mixed_precision_norm_01
 	add_config_var int_type_size
 	add_config_var blas_int_type_size
 	add_config_var enable_sup_handling       enable_sup_handling_01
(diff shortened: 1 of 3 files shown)


devinamatthews:
Could you move this into `bli_?normfv` and add a configure option to disable it (for those who *really* want the old behavior)?

EylonKrause:
Thanks for the review! I've updated the PR to address both points:

- **Moved into the kernel.** The double accumulation now lives in `bli_?normfv` (`frame/util/bli_util_unb_var1.c`) rather than the BLAS-compat layer, so both the native typed API (`bli_snormfv`/`bli_cnormfv`) and the BLAS/CBLAS `snrm2`/`scnrm2` benefit. The BLAS-layer change is reverted.
- **Configure option.** Added `--enable-mixed-precision-norm` (enabled by default); `--disable-mixed-precision-norm` restores the old single-precision accumulation. `dnrm2`/`dznrm2` are untouched and keep the scaled sum-of-squares path.

Verified on Zen4 (Ubuntu 24.04):
- default (enabled): native `bli_snormfv` and BLAS `snrm2` both return 30066.59 for `x=[30000, 4e6 ones]` (rel. error ~3e-8 vs a long double reference); `make checkblis-fast` = 2772/2772 PASS.
- `--disable-mixed-precision-norm`: reverts to 30000.0 (rel. error ~2e-3), confirming the toggle.

Implementation note: `normfv` has two scaled-sumsqv template instances (the live one and the `#else` of the disabled `FE_OVERFLOW`/dotv path), so the promotion hook is added to both. Happy to consolidate those or rename the option if you'd prefer.

configure
```diff
@@ -287,6 +287,13 @@ print_usage()
                  Enabling this option WILL NEGATIVELY IMPACT PERFORMANCE.
                  Please use only for informational/debugging purposes.
 
+   --enable-mixed-precision-norm, --disable-mixed-precision-norm
```
devinamatthews: Switch enable/disable order and start comment with "Disable..." c.f. `--disable-shared`.

frame/util/bli_util_unb_var1.c
```diff
@@ -265,6 +265,34 @@ void PASTEMAC(ch,varname) \
...
+// the sum of the squares in double precision. The sumsqv-based path below
+// otherwise accumulates in the working (single) precision, which is less
+// accurate than other BLAS implementations (MKL, OpenBLAS, vecLib). Overflow
+// is not a concern: a single-precision magnitude squared is at most ~1.16e77,
+// far below DBL_MAX (~1.8e308). See https://github.com/flame/blis/issues/914.
+// (Implemented as an outer helper macro because a macro body cannot itself
+// contain preprocessor directives.)
+#ifdef BLIS_ENABLE_MIXED_PRECISION_NORM
+#define BLIS_NORMFV_MIXED_PREC_PATH( ctype, ctype_r, ch, chr ) \
+	if ( sizeof( ctype_r ) == sizeof( float ) ) \
```
devinamatthews: Not super future-proof (e.g. int32?). Maybe:
```c
#define s_IS_FLOAT 1
#define d_IS_FLOAT 0
...
if ( PASTECH(chr,_IS_FLOAT) )
```

devinamatthews (dismissed):
Great work!

EylonKrause:
Thanks! Addressed both points:

- **`configure`** — reordered to `--disable-mixed-precision-norm, --enable-mixed-precision-norm` and reworded the help to start with "Disable (enabled by default)…", matching the `--disable-shared` convention.
- **Type check** — replaced the `sizeof( ctype_r ) == sizeof( float )` test with a per-real-char predicate: `#define s_IS_FLOAT 1` / `#define d_IS_FLOAT 0` and `if ( PASTECH(chr,_IS_FLOAT) )`, so the promotion keys on the precision domain rather than type size (future-proof against a same-size non-float real type). The two `s_IS_FLOAT`/`d_IS_FLOAT` defines are `#undef`'d after the `normfv` instantiations.

Rebuilt on Zen4 (Ubuntu 24.04): `make` clean, and the check still promotes correctly — `snrm2` on `x = [30000, 4e6 ones]` returns `30066.59` (rel. error 3.2e-8 vs a long-double reference), vs `30000.0` when built with `--disable-mixed-precision-norm`.

PR #935, opened 2026-07-01
bli_cpuid: clear feature bitfield through the pointer when OSXSAVE is absent

EylonKrause:
## Description

`bli_cpuid_query(uint32_t* family, uint32_t* model, uint32_t* features)` builds the CPU feature bitfield through the `features` pointer (`*features = 0` initially, `*features |= FEATURE_...` per detected feature).

The final guard handles the case where cpuid reports xsave/xgetbv is unsupported **or the OS hasn't enabled it** — in which case, per its own comment, "we clear the bitfield, because it means that not even xmm support is present." But it does:

```c
features = 0;   // nulls the local pointer (dead store) — the bitfield is NOT cleared
```

instead of `*features = 0;`. So when the OS lacks OSXSAVE, the SSE/AVX **hardware** bits stay set in the caller's bitfield even though the OS can't save/restore that vector register state. BLIS can then select a subconfig that issues instructions the OS can't handle → wrong dispatch, potentially `SIGILL`.

The sibling clear-site (the `!XGETBV_MASK_XMM` branch just above) and the initial clear both correctly use `*features = 0`.

## Fix

`features = 0;` → `*features = 0;`

## Testing

One-character pointer-dereference fix; correctness follows from the two sibling `*features = 0` sites and the branch's own comment. The trigger requires a host/VM that reports SSE/AVX in cpuid but doesn't enable OSXSAVE, so I couldn't exercise it on this machine.

### DIFF
--- a/frame/base/bli_cpuid.c
+++ b/frame/base/bli_cpuid.c
@@ -866,7 +866,7 @@ uint32_t bli_cpuid_query
 			// present.
 
 			//fprintf(stderr, "xgetbv: no\n");
-			features = 0;
+			*features = 0;
 		}
 	}
 
--- a/frame/base/bli_cpuid.c
+++ b/frame/base/bli_cpuid.c
@@ -862,25 +862,17 @@ uint32_t bli_cpuid_query
 		{
 			// If the hardware does not support xsave/xrestor/xsetbv/xgetbv,
 			// OR these features are not enabled by the OS, then the OS
-			// cannot manage the ymm (AVX) or zmm (AVX-512) register state,
-			// so we clear those feature bits. SSE-family features use the
-			// legacy xmm state (saved via FXSAVE, which does not depend on
-			// xsave/xgetbv), so we keep them: a processor may support
-			// additional SSE instruction sets without exposing
-			// xgetbv/osxsave, and CPUID can still be trusted for those.
+			// cannot manage the ymm (AVX) or zmm (AVX-512) register state.
+			// SSE-family features use the legacy xmm state (saved via FXSAVE,
+			// which does not depend on xsave/xgetbv), so we keep only those:
+			// a processor may support additional SSE instruction sets without
+			// exposing xgetbv/osxsave, and CPUID can still be trusted for them.
 
 			//fprintf(stderr, "xgetbv: no\n");
-			*features &= ~( FEATURE_AVX512F  |
-			                FEATURE_AVX512DQ |
-			                FEATURE_AVX512PF |
-			                FEATURE_AVX512ER |
-			                FEATURE_AVX512CD |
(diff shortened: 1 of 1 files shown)


devinamatthews:
Looks like making it work correctly actually fails the Penryn (e.g. Woodcrest) test because even without OSXSAVE the processor must support SSE and SSE2. AFAICT if a processor supports additional SSE instruction sets but not xgetbv/osxsave then we can trust cpuid. So, please keep any bits related to SSE instructions in that codepath.

EylonKrause:
Disclosure: this contribution was authored with an AI coding assistant (Claude) and reviewed before submission.

EylonKrause:
Good catch, thanks — you're right. SSE-family features (SSE, SSE2, and the SSE3/SSSE3/SSE4.1/SSE4.2 bits tracked here) use the legacy xmm state saved via FXSAVE, which doesn't depend on xgetbv/osxsave, so cpuid can be trusted for them even when xsave/osxsave is absent.

I updated the else branch to clear only the ymm/zmm-dependent bits (AVX, AVX2, FMA3, FMA4, and the AVX-512 set), mirroring the two guarded clears in the xgetbv-present path, and keep the SSE bits. That should let Penryn/Woodcrest retain their SSE feature bits — I believe it's what was failing `build-2`. Could you kick off CI to confirm?

frame/base/bli_cpuid.c
```diff
@@ -861,12 +861,26 @@ uint32_t bli_cpuid_query
...
+			// cannot manage the ymm (AVX) or zmm (AVX-512) register state,
+			// so we clear those feature bits. SSE-family features use the
+			// legacy xmm state (saved via FXSAVE, which does not depend on
+			// xsave/xgetbv), so we keep them: a processor may support
+			// additional SSE instruction sets without exposing
+			// xgetbv/osxsave, and CPUID can still be trusted for those.
 
 			//fprintf(stderr, "xgetbv: no\n");
-			features = 0;
+			*features &= ~( FEATURE_AVX512F  |
```
devinamatthews: For future-proofing, this should be all *SSE* flags, not inverse of AVX+AVX512.

devinamatthews (changes requested):

EylonKrause:
Good call — switched to masking the SSE-family bits directly (`*features &= (FEATURE_SSE3 | FEATURE_SSSE3 | FEATURE_SSE41 | FEATURE_SSE42)`) instead of clearing the inverse AVX/FMA/AVX-512 set. That way a feature bit added to the enum later can't be accidentally retained on a processor without xsave/xgetbv.

devinamatthews:
Thanks for making these changes.

PR #940, opened 2026-07-16
Add optional huge-page-backed memory pool

jeffhammond:
Back BLIS's internal memory pools (the packed A/B/C buffers streamed by level-3 ops) with explicit huge pages when the OS provides them, reducing dTLB pressure and page walks for large operands.

New allocator bli_hugepage_malloc / bli_hugepage_free (malloc/free- compatible) plug into BLIS_MALLOC_POOL / BLIS_FREE_POOL exactly like the existing libmemkind hbw_malloc hook. Per block, largest page first: 1 GiB hugetlb (MAP_HUGE_1GB) for >=1 GiB requests, then 2 MiB hugetlb (MAP_HUGE_2MB) for >=2 MiB, else ordinary malloc(). hugetlb blocks are mmap/munmap'd, malloc blocks free()'d; a small header before the returned pointer records which. The malloc fallback makes this a no-op when no hugetlb pool is reserved (and on a THP=always host the kernel promotes the malloc'd mmaps anyway), so there is no regression.

Opt-in via -DBLIS_ENABLE_HUGEPAGE_POOL (off by default, like memkind).

### DIFF
--- a/frame/base/bli_malloc.c
+++ b/frame/base/bli_malloc.c
@@ -38,6 +38,143 @@
 //#define BLIS_ENABLE_MEM_TRACING
 
 // -----------------------------------------------------------------------------
+// Huge-page-backed pool allocator.
+//
+// When BLIS_ENABLE_HUGEPAGE_POOL is defined, BLIS's internal memory pools
+// (used for the packed A/B/C buffers, which are the hot, repeatedly-streamed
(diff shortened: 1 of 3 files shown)


chillenb:
I think bli_fmalloc_align and bli_ffree_align should be rewritten too. And ideally each numa-node should get a separate pool..

PR #943, opened 2026-07-16
Multithread the memory-bound level-1/2 operations (OpenMP)

jeffhammond:
BLIS runs level-1 and level-2 single-threaded, so large vector/matrix-vector ops are capped at one core's memory bandwidth. Add optional OpenMP threading (opt-in via BLIS_ENABLE_L1_OPENMP) for the memory-bound ops:

  - level-1v: axpyv, dotv, scal2v, scalv/invscalv/setv, copyv/addv/subv (disjoint chunks; dotv uses a partial-sum-then-combine reduction);
  - util: asumv, and nrm2 (normfv) threaded while preserving its overflow-safe scaling via a LAPACK-style pairwise (scale,sumsq) combine;
  - level-2: gemv (both transposes -- row-parallel axpyf / output-parallel dotxf) and ger (both variants -- column/row parallel).

Only fires for large operands, when not already nested in a parallel region. Reduction combines are correct for real and complex types. Default builds (macro undefined) keep the original single-threaded behavior, so no regression. On a 10-core cluster this is a ~2-9x speedup for these ops (e.g. axpy 4->11, gemv 8->22 GFLOP/s, asum 15->90, nrm2 7->69 GB/s).

### DIFF
--- a/frame/1/bli_l1v_tapi.c
+++ b/frame/1/bli_l1v_tapi.c
@@ -36,6 +36,133 @@
 // #included from files that define the typed API macros.
 #ifdef BLIS_ENABLE_TAPI
 
+// -- Optional OpenMP threading for memory-bound level-1v operations ----------
+// BLIS runs level-1/2 single-threaded, which leaves most of the memory
+// bandwidth on the table for large vectors (a single GB10 X925 core sees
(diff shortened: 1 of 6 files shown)


chillenb:
Is this from NVPL BLAS? Cool!

PR #946, opened 2026-07-29
Add support for the s390x architecture 

jamesnohilly:
This PR introduces initial support for the s390x (z/Architecture) platform in BLIS. At present, architecture detection defaults to z15.

I also introduce two Vector-facility intrinsic-based microkernels, namely, `dgemm` (8x4) and `sgemm` (16x4).

If there are any further optimization strategies specific or styles to BLIS, feel free to suggest! I am happy to adapt the microkernel implementation with any suggestions.


jamesnohilly:
I have now added CI support for s390x through QEMU which uses Ubuntu 24.04.

PR #951, opened 2026-08-26
cblas: fix infinite loop in c/zgemv RowMajor ConjTrans with M==0

EylonKrause:
### Summary
`cblas_cgemv()` and `cblas_zgemv()` hang forever when called as
`cblas_?gemv(CblasRowMajor, CblasConjTrans, M = 0, N > 0, ...)`.

### Bug
In the `CblasRowMajor` + `CblasConjTrans` path, `Y` is conjugated only inside
`if (M > 0) { ... if (N > 0) { ... } }`, but the post-call restore loop that
un-conjugates `Y` is gated on `if (N > 0)` alone. When `M == 0` and `N > 0`,
the conjugation block is skipped, so the loop stride `i` and the sentinel `st`
are never assigned (they remain 0). The restore loop then executes

```c
do { *y = -(*y); y += i; } while (y != st);
```

with `i == 0` and `st == 0`, so `y` never advances and never equals `st` — an
infinite loop. A normal `M > 0` ConjTrans call is unaffected.

### Fix
Gate the un-conjugation on the same condition under which `Y` was conjugated
(`M > 0 && N > 0`), in both `cblas_cgemv.c` and `cblas_zgemv.c`.

### Verification
- Built with `./configure --enable-cblas auto && make` (config `zen3`, gcc 13.3).
- Reproducer `cblas_cgemv(CblasRowMajor, CblasConjTrans, 0, 2, ...)` under `timeout 10`:
  - against the **pristine** wrapper it times out (exit 124) — hang reproduced;
  - against the **fixed** wrapper it returns immediately (exit 0).
- `make checkblas` → `All BLAS tests passed!`

Disclosure: this contribution was authored with an AI coding assistant (Claude) and reviewed before submission.

### DIFF
--- a/frame/compat/cblas/src/cblas_cgbmv.c
+++ b/frame/compat/cblas/src/cblas_cgbmv.c
@@ -150,7 +150,7 @@ void cblas_cgbmv(enum CBLAS_ORDER order,
       if (TransA == CblasConjTrans)
       {
          if (x != X) free(x);
-         if (N > 0)
+         if (M > 0 && N > 0)
          {
             do
             {
--- a/frame/compat/cblas/src/cblas_cgemv.c
+++ b/frame/compat/cblas/src/cblas_cgemv.c
@@ -146,7 +146,7 @@ void cblas_cgemv(enum CBLAS_ORDER order,
       if (TransA == CblasConjTrans)
       {
          if (x != (const float *)X) free(x);
-         if (N > 0)
+         if (M > 0 && N > 0)
          {
             do
             {
--- a/frame/compat/cblas/src/cblas_zgbmv.c
+++ b/frame/compat/cblas/src/cblas_zgbmv.c
@@ -150,7 +150,7 @@ void cblas_zgbmv(enum CBLAS_ORDER order,
       if (TransA == CblasConjTrans)
       {
          if (x != X) free(x);
-         if (N > 0)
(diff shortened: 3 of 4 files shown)


devinamatthews:
You should also submit an issue/PR [here](https://github.com/Reference-LAPACK/lapack/), since we pulled this directly from netlib. It also seems to affect `cblas_[cz]gbmv`.

EylonKrause:
Thanks — good catch on `[cz]gbmv`. That path has the same `M == 0` early-out with the incrementing loop variable, so it almost certainly hangs the same way; I'll verify it and either fold it into this PR or send a follow-up. I'll also open the fix upstream at Reference-LAPACK since that's the source of truth here.

EylonKrause:
Done — folded the `[cz]gbmv` fix into this PR (2e581bf). Same one-liner: the restore loop is gated on `if (M > 0 && N > 0)` now, matching where `i`/`st` actually get assigned.

I reproduced it against a built libblis to be sure: `cblas_cgbmv(CblasRowMajor, CblasConjTrans, /*M*/0, /*N*/2, ...)` spins forever (killed by a 5s timeout) before the change and returns immediately with Y untouched after it.

I'll send the same fix (gemv + gbmv) upstream to Reference-LAPACK's CBLAS wrappers separately, since that's where it originates.

